PR #2, opened 2017-04-06
Basic Mailgun setup with Bamboo.

rhnonose:


rhnonose:
Better to just put in master.

config/prod.exs
```diff
@@ -14,3 +14,8 @@ config :course_planner, CoursePlanner.Repo,
   url: System.get_env("DATABASE_URL"),
   pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
   ssl: true
+
+config :course_planner, CoursePlanner.Mailer,
+  adapter: Bamboo.MailgunAdapter,
+  api_key: System.get_env("MAILGUN_API_KEY"),
+  domain: "sandbox2c32a5cb470c4c3480c227b8ed604f75.mailgun.org"
```
ggpasqualino: What do think of getting this value from the environment also?
If we have staging server for example, or we provide this as a self-hosted application, I suppose it would be better to have this coming from the env

rhnonose: I believe we can leave like that until we actually have a separate staging environment.
Also we can have staging.ex as a separate config and keep there, no need to have it in a environment variable.

ggpasqualino: I am not sure the config files should be used for data difference, this is usually a bad practice, they should specify behaviour differences and get the data from the environment

And there is still the case of offering this as a self hosted app, in this case it doesn't make sense to have this hardcoded at all

rhnonose: That makes sense, I'll change it.

PR #3, opened 2017-04-11
Setup coherence

rhnonose:
Setup [coherence](https://github.com/smpallen99/coherence) --full-invitation type.
Setup Swoosh with Mailgun.


config/config.exs
```diff
@@ -15,3 +15,17 @@ config :logger, :console,
...
+
+# %% Coherence Configuration %%   Don't remove this line
+config :coherence,
+  user_schema: CoursePlanner.User,
+  repo: CoursePlanner.Repo,
+  module: CoursePlanner,
+  logged_out_url: "/",
+  email_from_name: "Your Name",
+  email_from_email: "yourname@example.com",
+  opts: [:authenticatable, :recoverable, :lockable, :trackable, :unlockable_with_token, :invitable, :registerable]
```
ggpasqualino: @ghatighorias and I were talking about the users today, and they should not be `registerable`, it should be invite only, could you update it to be like that?

priv/repo/migrations/20170407143340_create_course.exs
```diff
@@ -0,0 +1,13 @@
+defmodule CoursePlanner.Repo.Migrations.CreateCourse do
```
ggpasqualino: Are courses necessary to setup coherence? Could you split it to another PR?

rhnonose: It's important to test if it's working properly to have a custom resource.

config/config.exs
rhnonose: It's weird, I had the idea that --full-invitable didn't include registerable. I'll check.

config/config.exs
```diff
@@ -15,3 +15,17 @@ config :logger, :console,
...
 import_config "#{Mix.env}.exs"
+
+# %% Coherence Configuration %%   Don't remove this line
+config :coherence,
+  user_schema: CoursePlanner.User,
+  repo: CoursePlanner.Repo,
+  module: CoursePlanner,
+  logged_out_url: "/",
+  email_from_name: "Your Name",
+  email_from_email: "yourname@example.com",
```
ggpasqualino: I think it would be good to get `email_from_name` and `email_from_email` from the environment

priv/repo/seeds.exs
```diff
@@ -1,11 +1,4 @@
...
-#     mix run priv/repo/seeds.exs
-#
-# Inside the script, you can read and write to any of your
-# repositories directly:
-#
-#     CoursePlanner.Repo.insert!(%CoursePlanner.SomeModel{})
-#
-# We recommend using the bang functions (`insert!`, `update!`
-# and so on) as they will fail if something goes wrong.
+CoursePlanner.Repo.delete_all CoursePlanner.User
```
rhnonose: Sometimes I pollute the database with further tests, so I just run the seeds again instead of dropping and setting up the database again. I wouldn't necessarily say it's good practice, but mostly common practice.
But maybe we shouldn't have this seed at all, since it's run in every environment. Do you know any way to run separate seeds for dev/test/prod?

config/config.exs
rhnonose: Done.

priv/repo/seeds.exs
ggpasqualino: I am note sure what is a good practice for seeds, but we can have `priv/repo/prod_seeds.exs` and `priv/repo/dev_seeds.exs`, then we interpolate the environment into the file name when we define the `ecto.setup` alias

rhnonose: I'm finding it difficult to separate into different files. I came up with a mix task, so it may be deemed refactoring, but according to [this issue](https://github.com/phoenixframework/phoenix_ecto/issues/8), this is a good middle-ground.

ggpasqualino: this doesn't seem like their final word in the subject though

> For those following at home, you can do: mix run priv/repo/seeds.exs in your Phoenix app and that's all. The seeds.exs file is full of comments to tell you what to do and what not to do.

rhnonose: Didn't find anything specifically helpful in the generated comments.

ggpasqualino: the main point is that you don't need a task, you can run `mix run priv/repo/seeds.exs` and `seeds.exs` is a plain elixir script

I suggest we leave it as you did before, let's keep the `seeds.exs`. I will check what to do about it

rhnonose: Alright, I'll revert it.

PR #7, opened 2017-04-16
Add docker setup

ggpasqualino:
Add docker setup to easy development onboarding

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,18 +3,10 @@ FROM elixir:1.4.2
 ENV NODE_VERSION=7
 
 # Install system dependencies and nodejs, then clean up apt temporary artefacts
-RUN apt-get -y update \
-    && apt-get -y install apt-transport-https build-essential curl git make locales locales-all inotify-tools \
-    && curl -sL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash - \
-    && apt-get -y install nodejs \
+RUN curl -sL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash - \
+    && apt-get -y install nodejs inotify-tools \
     && apt-get -y clean \
     && rm -rf /var/cache/apt/*
-
-# Elixir requires UTF-8
-RUN locale-gen en_US.UTF-8
-ENV LANG en_US.UTF-8
-ENV LC_ALL en_US.UTF-8
-ENV LANGUAGE en_US:en
 
 # Install Hex and Rebar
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN apt-get -y update \
+    && apt-get -y install apt-transport-https build-essential curl git make locales locales-all inotify-tools
+
+# Install nodejs
+RUN curl -sL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash - \
+    && apt-get -y install nodejs
+
+# Clean up apt temporary apt artefacts
+RUN apt-get clean -y \
+    && rm -rf /var/cache/apt/*
```
vbalazs: what is the intention of this command?

It will be a separate layer so the previous layers will still have the cache hence you don't win on the size of the image. For this to be useful you need to chain all the previous apt commands into one RUN.

README.md
```diff
@@ -2,7 +2,16 @@
...
 # Course Planner
 
-To start your Phoenix app:
+## Development Setup
+
+### Docker
+
+  * Install Elixir dependencies with `docker-compose run web mix deps.get`
+  * Setup your database (create, migrate and seed) with `docker-compose run web mix ecto.setup`
+  * Install Node.js dependencies with `docker-compose run web npm install`
```
vbalazs: I agree with @gdotdesign here, these should be build steps of the image (and they are already). If you change the dependencies you should do `docker-compose build` and docker should pick the changed file and rebuild from that layer.

If i understand it correctly that these deps are getting installed into the project's dir which gets overridden on volume mount. Is it possible to adjust the target directory of dependencies to something else?

edit: and i am not exactly sure how but this is not a problem with npm in `dh_alfred` image however i assume it uses the `node_modules` subdir. or i had the libs installed locally in the project folder too 😂  i can't recall

Dockerfile
ggpasqualino: Cool! Thanks!
And do you think it makes sense to join all the apt commands in favor of the container size? Or just remove this cleaning command and leave it as it is?

vbalazs: i think if they don't change too often it's nice if you chain them together. With separating into new lines how you did already 👍

README.md
ggpasqualino: so, in order to not change too much the project itself I changed the path to mount the project to a different one from the original of the image, then I start the docker-compose with a custom script which creates symbolic links to the download dependencies and compiled files

PR #11, opened 2017-04-19
Adds Swoosh Email integration

ghatighorias:


rhnonose:
Can you include the prod.exs configuration with environment variables that is already setup in master for swoosh/mailgun?

lib/course_planner.ex
```diff
@@ -1,20 +1,25 @@
 defmodule CoursePlanner do
+  @moduledoc """
+  This is the main module of the app
+  """
   use Application
+  alias CoursePlanner.Endpoint, as: Endpoint
+  alias CoursePlanner.Repo, as: Repo
```
rhnonose: you can shorten it with `alias CoursePlanner.{Endpoint, Repo}`
also, `alias CoursePlanner.Endpoint` and `CoursePlanner.Endpoint, as: Endpoint` is equivalent.

lib/course_planner/Mailer/main.ex
```diff
@@ -0,0 +1,6 @@
+defmodule CoursePlanner.Mailer.Main do
```
rhnonose: `CoursePlanner.Main.Mailer` or `CoursePlanner.Mailer` goes better with Elixir convention.

config/config.exs
```diff
@@ -14,6 +14,10 @@ config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
+config :course_planner, CoursePlanner.Mailer.Main,
+  adapter: Swoosh.Adapters.SMTP,
+  api_key: "some api key should be added here"
+
```
rhnonose: You can use `adapter: Swoosh.Adapters.Local` to check sent e-mails in `/dev/mailbox` and put the SMTP config in `prod.exs`.

lib/course_planner/Mailer/main.ex
ghatighorias: I named it based on folder structure. if I change Main and Mailer then it does not represent the structure. do you know a better name instead of Main ?

rhnonose: Maybe just leave it as `CoursePlanner.Mailer`
The other mailer is `CoursePlanner.Coherence.Mailer` which is specific to coherence, or change the folder structure.
Folder naming should be downcased as well.

ghatighorias: hmm, i prefer changing the folder structure, but rather needs time. so i'll think about it

config/prod.exs
```diff
@@ -23,3 +23,7 @@ config :coherence, CoursePlanner.Coherence.Mailer,
 config :coherence,
   email_from_name: System.get_env("EMAIL_FROM_NAME"),
   email_from_email: System.get_env("EMAIL_FROM_EMAIL")
+
+config :course_planner, CoursePlanner.Mailer.Main,
+  adapter: Swoosh.Adapters.SMTP,
```
ggpasqualino: I think we should drop smtp setup for now and use mailgun instead, like @rhnonose did with the coherence on line 18

PR #12, opened 2017-04-19
Adds a generic Enum type for Status on the database level

ghatighorias:


ggpasqualino:
I think you could add the ecto type in this PR too

ghatighorias:
@ggpasqualino I am not sure if I get your comment _I think you could add the ecto type in this PR too_

ggpasqualino:
@ghatighorias You showed me a custom Ecto.Type which you created, that will validate if the field is one of those statuses, you should add that type in this pr here too

mix.lock
```diff
@@ -12,6 +12,7 @@
   "dogma": {:hex, :dogma, "0.1.14", "561abecb25a2408738e7d2bc34cc565a07e6fadbc4fb91b2cbb35a5a9234cf1c", [:mix], [{:poison, ">= 2.0.0", [hex: :poison, optional: false]}]},
   "ecto": {:hex, :ecto, "2.1.4", "d1ba932813ec0e0d9db481ef2c17777f1cefb11fc90fa7c142ff354972dfba7e", [:mix], [{:db_connection, "~> 1.1", [hex: :db_connection, optional: true]}, {:decimal, "~> 1.2", [hex: :decimal, optional: false]}, {:mariaex, "~> 0.8.0", [hex: :mariaex, optional: true]}, {:poison, "~> 2.2 or ~> 3.0", [hex: :poison, optional: true]}, {:poolboy, "~> 1.5", [hex: :poolboy, optional: false]}, {:postgrex, "~> 0.13.0", [hex: :postgrex, optional: true]}, {:sbroker, "~> 1.0", [hex: :sbroker, optional: true]}]},
   "fs": {:hex, :fs, "0.9.2", "ed17036c26c3f70ac49781ed9220a50c36775c6ca2cf8182d123b6566e49ec59", [:rebar], []},
+  "gen_smtp": {:hex, :gen_smtp, "0.11.0", "d90ff2f021fc86cb2a4259b1f2b177ab6e506676265e26454bf5755855adc956", [:rebar3], []},
```
ggpasqualino: can you remove this change from this PR?

priv/repo/migrations/20170419143747_create_entity_status.exs
```diff
@@ -0,0 +1,16 @@
+defmodule CoursePlanner.Repo.Migrations.CreateEntityStatus do
+  use Ecto.Migration
+
+  def change do
+    execute("""
+          CREATE TYPE entity_status AS ENUM (
+            'Planned',
+            'Active',
+            'Finished',
+            'Graduated',
```
ggpasqualino: I thought `Graduated` wouldn't be necessary because we could consider a `Finished` student as `Graduated`

web/models/types/entity_type.ex
```diff
@@ -0,0 +1,27 @@
+defmodule CoursePlanner.Types.EntityType do
+  @moduledoc """
+    This module introduces a custom type for Etco for checking status in the model
+  """
+  @behaviour Ecto.Type
+  def type, do: :string
```
ggpasqualino: shouldn't this type be `entity_status` like you just created in the database?

rhnonose: no, this is the callback from the `Ecto.Type` behaviour

web/models/types/entity_type.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  @valid_entity_types ['Planned', 'Active', 'Finished', 'Graduated', 'Frozen', 'Deleted']
```
rhnonose: Shouldn't this be a string instead of a charlist? Did you test it?

web/models/types/entity_type.ex
ggpasqualino: sure, bur the callback docs say
> Returns the underlying schema type for the custom type.
> Note this function is not required to return Ecto primitive types, the type is only required to be known by the adapter.

Isn't this the db schema type?

web/models/types/entity_type.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  def valid_entity_types, do: @valid_entity_types
+
+  def cast(value) do
```
ggpasqualino: the callbacks `cast/1`, `dump/1`, and `load/1` should return `{:ok, term} | :error` according to the docs, right now you are returning `term | :error`

web/models/types/entity_type.ex
```diff
@@ -0,0 +1,27 @@
+defmodule CoursePlanner.Types.EntityType do
+  @moduledoc """
+    This module introduces a custom type for Etco for checking status in the model
+  """
+  @behaviour Ecto.Type
+  def type, do: :string
+
+  @valid_entity_types ['Planned', 'Active', 'Finished', 'Graduated', 'Frozen', 'Deleted']
+
+  def valid_entity_types, do: @valid_entity_types
```
ggpasqualino: fyi, he explained to me the idea of this function is to provide possible values to the views

web/models/types/entity_type.ex
rhnonose: Maybe you're right. I can't find a definite response on what I've found. Ultimately the value is a string, so not sure.
@ghatighorias does it work like this?

PR #13, opened 2017-04-21
Adds administration for courses

ghatighorias:


ghatighorias:
I'll fix the failing tests and commit again

web/router.ex
```diff
@@ -40,6 +40,9 @@ defmodule CoursePlanner.Router do
     pipe_through :protected
 
     resources "/users", UserController
+    resources "/courses", CourseController
+    resources "/courses/:id", CourseController
+    resources "/courses/:id/edit", CourseController
```
rhnonose: `resources "/courses", CourseController` does all you need.

mix.lock
```diff
@@ -12,6 +12,7 @@
   "dogma": {:hex, :dogma, "0.1.14", "561abecb25a2408738e7d2bc34cc565a07e6fadbc4fb91b2cbb35a5a9234cf1c", [:mix], [{:poison, ">= 2.0.0", [hex: :poison, optional: false]}]},
   "ecto": {:hex, :ecto, "2.1.4", "d1ba932813ec0e0d9db481ef2c17777f1cefb11fc90fa7c142ff354972dfba7e", [:mix], [{:db_connection, "~> 1.1", [hex: :db_connection, optional: true]}, {:decimal, "~> 1.2", [hex: :decimal, optional: false]}, {:mariaex, "~> 0.8.0", [hex: :mariaex, optional: true]}, {:poison, "~> 2.2 or ~> 3.0", [hex: :poison, optional: true]}, {:poolboy, "~> 1.5", [hex: :poolboy, optional: false]}, {:postgrex, "~> 0.13.0", [hex: :postgrex, optional: true]}, {:sbroker, "~> 1.0", [hex: :sbroker, optional: true]}]},
   "fs": {:hex, :fs, "0.9.2", "ed17036c26c3f70ac49781ed9220a50c36775c6ca2cf8182d123b6566e49ec59", [:rebar], []},
+  "gen_smtp": {:hex, :gen_smtp, "0.11.0", "d90ff2f021fc86cb2a4259b1f2b177ab6e506676265e26454bf5755855adc956", [:rebar3], []},
```
ggpasqualino: could you remove this? we don't require this package

maybe it is worth it to clean your deps, this is always coming on your PRs

web/controllers/course_controller.ex
```diff
@@ -0,0 +1,65 @@
...
+        render(conn, "edit.html", course: course, changeset: changeset)
+    end
+  end
+
+  def delete(conn, %{"id" => id}) do
+    course = Repo.get!(Course, id)
+
+    # Here we use delete! (with a bang) because we expect
+    # it to always work (and if it does not, it will raise).
+    Repo.delete!(course)
```
ggpasqualino: We should not delete them for real, we agreed to soft delete things

web/controllers/course_controller.ex
```diff
@@ -0,0 +1,65 @@
+defmodule CoursePlanner.CourseController do
+  use CoursePlanner.Web, :controller
+
+  alias CoursePlanner.Course
+
+  def index(conn, _params) do
+    courses = Repo.all(Course)
```
ggpasqualino: We should list only the not soft deleted things

web/controllers/course_controller.ex
ghatighorias: for this, shall I just set the status to deleted ?
and then in the list to show only those that are not having "deleted" as status ?

priv/repo/migrations/20170421092531_create_course.exs
```diff
@@ -0,0 +1,17 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:courses) do
+      add :name, :string
+      add :description, :string
+      add :number_of_sessions, :integer
+      add :session_duration, :time
+      add :syllabus, :string
+      add :status, :entity_status
```
ggpasqualino: we should have a `deleted_at` timestamp also

web/models/course/course.ex
```diff
@@ -0,0 +1,32 @@
...
+    timestamps()
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:name, :description, :number_of_sessions, :session_duration, :syllabus, :status])
+    |> validate_required([:name, :description, :number_of_sessions, :session_duration, :syllabus, :status])
```
ggpasqualino: is `syllabus` really required? I thought it was optional

web/models/course/course_helper.ex
```diff
@@ -0,0 +1,31 @@
...
+  @moduledoc """
+  This module provides custom functionality for controller over the model
+  """
+  use CoursePlanner.Web, :model
+  import Ecto.DateTime, only: [utc: 0]
+
+  alias CoursePlanner.Repo
+  alias CoursePlanner.Course
+
+  def delete_handler(course) do
```
ggpasqualino: what do you think of renaming this function to `delete`?

web/models/course/course_helper.ex
```diff
@@ -0,0 +1,31 @@
...
+    case course.status do
+      "Planned" -> hard_delete_course(course)
+      _         -> soft_delete_course(course)
+    end
+  end
+
+  defp soft_delete_course(course) do
+    changeset = change(course, %{status: "Deleted", deleted_at: utc()})
```
ggpasqualino: I gave some thought to this `Deleted` status and I think we should remove it, the status is something meaningful to the user and we are overwriting it when we delete the record, so if we happen to need to recover a record we will lose this information.

What do you think of using only the `deleted_at` timestamp as part of soft-deleting records?

priv/repo/migrations/20170421092531_create_course.exs
ghatighorias: I was thinking since we have updated_at and delete status it works the same way. but if i can drop delete status and add "deleted_at".

However I have doubt now in general of to have soft delete in any case.

web/models/course/course_helper.ex
ghatighorias: As I wrote above, now i have big doubt wether a deleted item should be at all kept. since we have no plan of giving access to the deleted data. plus is not so informative. we can even make it easier by generally dropping soft delete and just hard delete any item

test/models/course_test.exs
```diff
@@ -0,0 +1,93 @@
...
+    assert changeset.valid?
+  end
+
+  test "changeset with status Frozen" do
+    changeset = Course.changeset(%Course{}, %{ @valid_attrs | status: "Frozen" })
+    assert changeset.valid?
+  end
+
+  test "changeset with invalid status with :create" do
+    changeset = Course.changeset(%Course{}, %{ @valid_attrs | status: "random" })
```
ggpasqualino: do you mean `Course.changeset(%Course{}, %{ @valid_attrs | status: "random" }, :create)` ?

web/controllers/course_controller.ex
```diff
@@ -0,0 +1,66 @@
...
+        |> redirect(to: course_path(conn, :show, course))
+      {:error, changeset} ->
+        render(conn, "edit.html", course: course, changeset: changeset)
+    end
+  end
+
+  def delete(conn, %{"id" => id}) do
+    course = Repo.get!(Course, id)
+
+    # Here we use delete! (with a bang) because we expect
```
ggpasqualino: could you remove this comment?

test/models/course_test.exs
ghatighorias: I have two changeset in the model, one general and one specific to create. so I wanted to test both in the model test file. check /models/course/course.ex

web/controllers/course_controller.ex
ghatighorias: yep. sorry it's a left over from the orig code

test/models/course_test.exs
ggpasqualino: yes, I got it, and that's why I am commenting
you didn't do what you intended to do, you test twice the following
```elixir
Course.changeset(%Course{}, %{ @valid_attrs | status: "random" })
```
instead of testing once
```elixir
Course.changeset(%Course{}, %{ @valid_attrs | status: "random" })
```
and once
```elixir
Course.changeset(%Course{}, %{ @valid_attrs | status: "random" }, :create)
```

ghatighorias: Truuuuuueeeee. such stupid mistake. I'll fix it

PR #18, opened 2017-04-24
Add user role.

rhnonose:
Add user role through the create user form.


priv/repo/migrations/20170424080542_create_role.exs
```diff
@@ -0,0 +1,21 @@
+defmodule CoursePlanner.Repo.Migrations.CreateRole do
+  use Ecto.Migration
+  alias CoursePlanner.Types.UserRole
+
+  def change do
+    execute("""
+          CREATE TYPE user_role AS ENUM (
+            'Student',
+            'Teacher',
+            'Organizer',
```
ggpasqualino: Organizer should be coordinator now

PR #19, opened 2017-04-24
Delete user

rhnonose:
Soft delete user and not show in the index.

### DIFF
--- a/lib/user_role.ex
+++ b/lib/user_role.ex
@@ -1,0 +1,27 @@
+defmodule CoursePlanner.Types.UserRole do
+  @moduledoc """
+    This module introduces a custom type for Etco for checking user role in the user model
+  """
+  @behaviour Ecto.Type
+  def type, do: :user_role
+
+  @valid_values ["Student", "Teacher", "Organizer"]
+
+  def values, do: @valid_values
+
+  def cast(value) do
+    case Enum.member?(@valid_values, value) do
+      :true  -> {:ok, value}
+      :false -> :error
+    end
+  end
+
+  def load(value), do: {:ok, value}
+
+  def dump(value) do
+    case Enum.member?(@valid_values, value) do
+      :true  -> {:ok, value}
+      :false -> :error
+    end
+  end
+end
--- a/priv/repo/migrations/20170424091310_soft_delete_user.exs
+++ b/priv/repo/migrations/20170424091310_soft_delete_user.exs
@@ -1,0 +1,10 @@
+defmodule CoursePlanner.Repo.Migrations.SoftDeleteUser do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :deleted, :boolean
+      add :deleted_at, :utc_datetime
+    end
+  end
+end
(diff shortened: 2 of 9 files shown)


lib/user_role.ex
```diff
@@ -0,0 +1,27 @@
+defmodule CoursePlanner.Types.UserRole do
+  @moduledoc """
+    This module introduces a custom type for Etco for checking user role in the user model
+  """
+  @behaviour Ecto.Type
+  def type, do: :user_role
+
+  @valid_values ["Student", "Teacher", "Organizer"]
```
ghatighorias: we need "volenteer" too :)

rhnonose: right :)

ghatighorias:
Quick question : is it possible to remove all the organizer users ?
cause if so then it will be problematic ( an app which won't have any admin )

rhnonose:
Maybe we can have an admin user as backup, but I don't think it's likely when we setup the permissions.

priv/static/js/app.js
```diff
@@ -1549,8 +1549,8 @@ channel.join().receive("ok", function (resp) {
 exports.default = socket;
 });
 
-;require.alias("phoenix_html/priv/static/phoenix_html.js", "phoenix_html");
-require.alias("phoenix/priv/static/phoenix.js", "phoenix");require.register("___globals___", function(exports, require, module) {
+;require.alias("phoenix/priv/static/phoenix.js", "phoenix");
```
ggpasqualino: can we ignore this file? it keeps coming and going when it is regenerated

rhnonose: Yeah, I still don't know what's the problem.

priv/repo/migrations/20170424091310_soft_delete_user.exs
```diff
@@ -0,0 +1,10 @@
+defmodule CoursePlanner.Repo.Migrations.SoftDeleteUser do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :deleted, :boolean
+      add :deleted_at, :utc_datetime
```
rhnonose: Good idea.

priv/repo/migrations/20170424091310_soft_delete_user.exs
```diff
@@ -0,0 +1,9 @@
+defmodule CoursePlanner.Repo.Migrations.SoftDeleteUser do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :deleted_at, :utc_datetime
```
ggpasqualino: I was talking to @ghatighorias about this type. What do you think of using `naive_datetime`?
The standard timestamps (inserted_at, updated_at) use this type

rhnonose: I don't mind. Phoenix is going to support utc_datetime generators, but the standard will be naive_datetime.
Unless we're going to deal with timezones, we can set all to naive_datetime.

ggpasqualino: okay, I say that we make it naive then, to keep it consistent with the others

rhnonose: One issue though, coherence has utc_datetime types.
The effort to keep it consistent should be higher for naive_datetime.

ggpasqualino:
can you create a new pr without this user roles things in it? 
github doesn't seem to get the diff properly after I merged the other pr about roles

PR #21, opened 2017-04-25
List all students

rhnonose:
I created a separate controller/view/template for the students to not clump the user controller too much and to separate the permissions as well when it comes to placing it, but I'm not too sure about it.
Thoughts @ggpasqualino @ghatighorias ?

### DIFF
--- a/web/controllers/student_controller.ex
+++ b/web/controllers/student_controller.ex
@@ -1,0 +1,12 @@
+defmodule CoursePlanner.StudentController do
+  use CoursePlanner.Web, :controller
+  alias CoursePlanner.User
+  alias CoursePlanner.Router.Helpers
+  import Ecto.Query
+
+  def index(conn, _params) do
+    query = from u in User, where: u.role == "Student"
+    students = Repo.all(query)
+    render(conn, "index.html", students: students)
+  end
+end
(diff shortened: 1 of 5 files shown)


ggpasqualino:
Having a separated controller/view/template is very reasonable to me 👍

web/controllers/student_controller.ex
```diff
@@ -0,0 +1,12 @@
+defmodule CoursePlanner.StudentController do
+  use CoursePlanner.Web, :controller
+  alias CoursePlanner.User
+  alias CoursePlanner.Router.Helpers
+  import Ecto.Query
+
+  def index(conn, _params) do
+    query = from u in User, where: u.role == "Student"
+    students = Repo.all(query)
```
ggpasqualino: you are missing the `deleted_at` condition here

ghatighorias:
As I can see you don't show the status of the students. but in general . it's not a bad idea to have it there

web/controllers/student_controller.ex
rhnonose: Yeah right.

rhnonose:
I'll include the student status when putting the status change actions.

PR #23, opened 2017-04-25
Add term show page

ggpasqualino:


test/controllers/term_controller_test.exs
```diff
@@ -46,4 +46,26 @@ defmodule CoursePlanner.TermControllerTest do
...
+  test "doesn't show inexisting term", %{conn: conn} do
+    conn = get conn, term_path(conn, :show, 1)
+    assert html_response(conn, 404)
+  end
+
+  test "doesn't show deleted term", %{conn: conn} do
+    {:ok, term} = CoursePlanner.Terms.create_term(%{name: "Spring", start_date: "2017-04-25", end_date: "2017-05-25", status: "Planned"})
+    term
+    |> Term.changeset()
+    |> Ecto.Changeset.put_change(:deleted_at, Ecto.DateTime.utc())
```
rhnonose: Can you put this inside `terms.ex`? In case we remove the soft delete logic and the logic doesn't get scattered.

ggpasqualino: sure, however I would prefer to refactor this when I do the delete endpoint

PR #25, opened 2017-04-26
Add edit and update term actions

ggpasqualino:


web/controllers/term_controller.ex
```diff
@@ -29,9 +29,35 @@ defmodule CoursePlanner.TermController do
     end
   end
 
+  def edit(conn, %{"id" => id}) do
+    case Terms.edit(id) do
+      nil ->
```
rhnonose: Can you match on  something like `{:error, :not_found}` like in the update action?

PR #27, opened 2017-04-26
Edit students

rhnonose:
Edit student details.

### DIFF
--- a/web/models/students.ex
+++ b/web/models/students.ex
@@ -1,4 +1,7 @@
 defmodule CoursePlanner.Students do
+  @moduledoc """
+    Handle Students specific logics
+  """
   alias CoursePlanner.Repo
   alias CoursePlanner.User
   import Ecto.Query
@@ -6,7 +9,7 @@ defmodule CoursePlanner.Students do
 
   @students from u in User, where: u.role == "Student" and is_nil(u.deleted_at)
 
-  def all() do
+  def all do
     Repo.all(@students)
   end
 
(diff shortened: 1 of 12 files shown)


ggpasqualino:
can you add this [fix](https://github.com/digitalnatives/course_planner/pull/30/commits/2188ce7ff8505c84f739e548ba73e9a72f564175) to the current pr?

web/models/students.ex
```diff
@@ -0,0 +1,45 @@
...
+  def all do
+    Repo.all(@students)
+  end
+
+  def update(id, params) do
+    case Repo.get(User, id) do
+      student ->
+        student
+        |> User.changeset(params, :update)
+        |> add_timestamps()
```
ggpasqualino: I created a function for this `CoursePlanner.Statuses.update_status_timestamp/1`, but it requires changes in the names of the timestamps

rhnonose: I'll refactor later when trying to come up with a more generic enum type.

ggpasqualino: I don't see the relation between the two things. Can you elaborate?

rhnonose: You use `EntityStatus.values` directly, so it can be made generic for other Enums by passing the values as a parameter, for example.
Including this I might organize the code better and put in `lib`, since there's some other stuff inside `web`

ggpasqualino: okay, so using `CoursePlanner.Statuses` right now would make it easier to find the places to refactor I suppose

rhnonose: Doing that would require me to refactor the part of the code that I'm not touching, hence do it later when other things are merged.

PR #29, opened 2017-04-27
List all teachers

rhnonose:


ghatighorias:
I am not sure teachers had nickname. can you double check ?

ggpasqualino:
@ghatighorias teacher and student use the same database table for now (users table), so they will share this field

what we can do however is to have different changesets for each role

rhnonose:
We were discussing about making different forms for new users (now it's generic with a dropdown), so for now some fields might show up on each other.
It needs a bit of refactoring/reorganization in the user controller to reuse some of the coherence logic.

PR #31, opened 2017-04-28
Introduces class table, model and tests

ghatighorias:


priv/repo/migrations/20170427090331_create_class.exs
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:classes) do
+      add :class_date, :date
+      add :starting_at, :time
+      add :finishes_at, :time
+      add :status, :entity_status
+      add :deleted_at, :naive_datetime
+      add :course_id, references(:courses, on_delete: :delete_all)
```
rhnonose: Do you need `:delete_all` for this? Since we're not hard deleting the record, this could mess up things in the future.

ggpasqualino: I think that's fine as a database constraint, if we do happen to hard delete a course we don't leave orphan class records

priv/repo/migrations/20170427090331_create_class.exs
```diff
@@ -0,0 +1,18 @@
+defmodule CoursePlanner.Repo.Migrations.CreateClass do
+  use Ecto.Migration
+
+  def change do
+    create table(:classes) do
+      add :class_date, :date
```
ggpasqualino: what do you think of renaming this to `date`? this is already part of the class table

priv/repo/migrations/20170427090331_create_class.exs
ghatighorias: i was thinking the same cause in any case if the course is in planning stage it should be hard deleted

PR #32, opened 2017-04-28
Manage coordinator

rhnonose:
List, show and edit coordinator.
There's a few modifications from other branches that is polluting the PR, so I'll leave it here until it is merged and update it.
Does is sound good or should I wait until merged to create the PR, @ggpasqualino ?

### DIFF
--- a/lib/course_planner/users.ex
+++ b/lib/course_planner/users.ex
@@ -10,7 +10,7 @@ defmodule CoursePlanner.Users do
     %User{}
     |> User.changeset(user)
     |> Changeset.put_change(:reset_password_token, token)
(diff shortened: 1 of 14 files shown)


lib/course_planner/users.ex
```diff
@@ -0,0 +1,35 @@
...
+  """
+  alias CoursePlanner.Repo
+  alias CoursePlanner.User
+  alias Ecto.{Changeset, DateTime}
+
+  def new_user(user, token) do
+    %User{}
+    |> User.changeset(user)
+    |> Changeset.put_change(:reset_password_token, token)
+    |> Changeset.put_change(:reset_password_sent_at, DateTime.utc)
```
rhnonose: should indeed

PR #34, opened 2017-05-03
Create specific views for roles

rhnonose:
Create separate new and delete actions for the roles.


rhnonose:
It's more like reorganising the templates to be separated.
We'll deal with access levels later, but it's mostly for the coordinator to manage users by role.

PR #35, opened 2017-05-03
Manage volunteer

rhnonose:
List, show, edit, create and delete volunteer.

### DIFF
--- a/web/templates/volunteer/form.html.eex
+++ b/web/templates/volunteer/form.html.eex
@@ -1,0 +1,41 @@
+<%= form_for @changeset, @action, fn f -> %>
+  <%= if @changeset.action do %>
+    <div class="alert alert-danger">
+      <p>Oops, something went wrong! Please check the errors below.</p>
+    </div>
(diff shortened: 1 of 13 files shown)


web/templates/volunteer/form.html.eex
```diff
@@ -0,0 +1,41 @@
...
+
+  <div class="form-group">
+    <%= label f, :nickname, class: "control-label" %>
+    <%= text_input f, :nickname, class: "form-control" %>
+    <%= error_tag f, :nickname %>
+  </div>
+
+  <div class="form-group">
+    <%= label f, :email, class: "control-label" %>
+    <%= text_input f, :email, class: "form-control" %>
```
ghatighorias: dont you wanna use email_input(form, field, opts \\ []) instead of text_input ?

rhnonose: Not sure, I'll take a look, but will leave out of this PR since there's more places without it.

PR #36, opened 2017-05-04
Generalize Enum type

rhnonose:
Define a behaviour to help using Enum types easier, with an optional helper with the timestamp.
As you can see in `EntityStatus`, it's cleaner to define an Enum to be used. The schema type could still be an `EntityStatus` but a subset could be passed to restrict the choosable types (see `CoursePlanner.Students` and `CoursePlanner.StudentStatus`).

### DIFF
--- a/lib/enum_timestamp.ex
+++ b/lib/enum_timestamp.ex
@@ -2,6 +2,25 @@ defmodule EnumTimestamp do
   @moduledoc """
   Behaviour to represent Enum with mapping to timestamp
   """
-  @callback valid_types() :: [String.t]
-  @callback types_timestamp() :: [{String.t, atom()}]
+  defmacro __using__(_) do
+    quote do
+      @behaviour Ecto.Type
+      @behaviour EnumTimestamp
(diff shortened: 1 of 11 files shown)


lib/enum_timestamp.ex
```diff
@@ -0,0 +1,26 @@
+defmodule EnumTimestamp do
```
ggpasqualino: what do you think about renaming this to `CoursePlanner.Enum` or something else?  I don't think the focus should be on the timestamp, that's just some convenient information

lib/enum_timestamp.ex
```diff
@@ -0,0 +1,26 @@
...
+  @moduledoc """
+  Behaviour to represent Enum with mapping to timestamp
+  """
+  defmacro __using__(_) do
```
ggpasqualino: 👍

lib/enum_timestamp.ex
rhnonose: Yeah, this naming kind of sucks.

PR #39, opened 2017-05-08
Adds delete functionality for a course

ghatighorias:


web/controllers/class_controller.ex
```diff
@@ -51,12 +51,12 @@ defmodule CoursePlanner.ClassController do
...
-#    class = Repo.get!(Class, id)
-#    ClassHelper.delete(class)
-
-#    conn
-#    |> put_flash(:info, "Class deleted successfully.")
-#    |> redirect(to: class_path(conn, :index))
-#  end
+  def delete(conn, %{"id" => id}) do
+    class = Repo.get!(Class, id)
+    ClassHelper.delete(class)
```
rhnonose: Can you check the return for successful delete?

test/controllers/class_controller_test.exs
```diff
@@ -221,10 +221,52 @@ defmodule CoursePlanner.ClassControllerTest do
...
+  test "hard deletes chosen resource when status is Planned", %{conn: conn} do
+    {:ok, created_course} = create_course()
+    class_args = %Class{course_id: created_course.id, date: Ecto.Date.from_erl({2010, 01, 01}), starting_at: Ecto.Time.from_erl({13, 0, 0}), finishes_at: Ecto.Time.from_erl({14, 0, 0}), status: "Planned"}
+    class = Repo.insert! class_args
+    conn = delete conn, class_path(conn, :delete, class)
+    assert redirected_to(conn) == class_path(conn, :index)
+    refute Repo.get(Class, class.id)
+  end
+
+  test "soft deletes chosen resource when status is Active", %{conn: conn} do
```
rhnonose: I don't think it's necessary to test all status for the delete action.
If you wish so, try to make a generic code so it's not too verbose, like:
```elixir
  Enum.map ["Active", "Frozen"], fn(status) ->
    test "soft deletes chosen resource when status is #{status}", %{conn: conn} do
      #your test here
    end
  end
```

web/controllers/class_controller.ex
```diff
@@ -51,12 +51,12 @@ defmodule CoursePlanner.ClassController do
...
-#  def delete(conn, %{"id" => id}) do
-#    class = Repo.get!(Class, id)
-#    ClassHelper.delete(class)
-
-#    conn
-#    |> put_flash(:info, "Class deleted successfully.")
-#    |> redirect(to: class_path(conn, :index))
-#  end
+  def delete(conn, %{"id" => id}) do
+    class = Repo.get!(Class, id)
```
rhnonose: You can put the get inside the `ClassHelper` module to remove the `Repo` dependency from the controller.

test/controllers/class_controller_test.exs
```diff
@@ -221,10 +221,30 @@ defmodule CoursePlanner.ClassControllerTest do
     assert html_response(conn, 200) =~ "Edit class"
```
ggpasqualino: could you add a test for deleting a non-existent id?

ghatighorias: done

PR #41, opened 2017-05-10
Add Term's holidays nested fields

ggpasqualino:
* Add dynamic fields for holidays, add/remove fields on the fly
* Add empty list for holidays if there is no holidays parameter when
  updating a term
* Replace the whole holidays json when saving


ggpasqualino:
@stfbax

PR #44, opened 2017-05-10
Manage tasks

rhnonose:


web/templates/task/index.html.eex
```diff
@@ -0,0 +1,33 @@
...
+      <th>Assigned to</th>
+    </tr>
+  </thead>
+  <tbody>
+<%= for task <- @tasks do %>
+    <tr>
+      <td><%= task.name %></td>
+      <td><%= task.deadline %></td>
+      <td><%= task.status %></td>
+      <%= if task.user != nil do %>
```
ggpasqualino: I am not sure which one is the most used in Elixir, but `if task.user != nil do` is equivalent to `if task.user do`.

Coming from ruby the latter is more common

ggpasqualino: What do you think of extracting this into a function in TaskView? it seems to repeat later

priv/repo/migrations/20170504190323_create_tasks.exs
```diff
@@ -0,0 +1,24 @@
...
+    execute("""
+          CREATE TYPE task_status AS ENUM (
+            'Pending',
+            'Accomplished'
+          )
+    """)
+    create table(:tasks) do
+      add :name, :string
+      add :deadline, :date
+      add :status, :task_status
```
ggpasqualino: I am also not sure about this status, if the task is not something to be delivered, maybe those statuses don't 
make sense

rhnonose: I'll leave it to be clarified by @ghatighorias then.

priv/repo/migrations/20170504190323_create_tasks.exs
```diff
@@ -0,0 +1,24 @@
...
+  def change do
+    execute("""
+          CREATE TYPE task_status AS ENUM (
+            'Pending',
+            'Accomplished'
+          )
+    """)
+    create table(:tasks) do
+      add :name, :string
+      add :deadline, :date
```
rhnonose: I'll leave it to be clarified by @ghatighorias then.

web/templates/task/index.html.eex
rhnonose: Makes sense.
I'll probably change to `if task.user_id` as well because `task.user` is filled with an ecto association, so I think the correct way is to check `task.user_id`.

priv/repo/migrations/20170504190323_create_tasks.exs
ggpasqualino: @ghatighorias ping

priv/repo/migrations/20170504190323_create_tasks.exs
ghatighorias: delivery of a task is not important. simply if it is assigned and informed is enough.

priv/repo/migrations/20170504190323_create_tasks.exs
ghatighorias: so far what we see as a task is something that should be done on a date and a period of time. is not important to give a deadline. but rather start and finish. think about it like this "i am available from 1999.1.1 10am till 11am to print course materials

priv/repo/migrations/20170504190323_create_tasks.exs
ghatighorias: I replied to the same somewhere else. status is not important cause tracking volunteers is way more simple, by just seeing whether they do it or not. i think we can drop the status

priv/repo/migrations/20170504190323_create_tasks.exs
rhnonose: I'll make the adjustments.

priv/repo/migrations/20170504190323_create_tasks.exs
```diff
@@ -0,0 +1,18 @@
...
+  use Ecto.Migration
+
+  def change do
+    create table(:tasks) do
+      add :name, :string
+      add :start_time, :naive_datetime
+      add :finish_time, :naive_datetime
+      add :user_id, references(:users)
+
+      add :pending_at, :naive_datetime
```
ggpasqualino: I don't think we need the timestamps `pending_at` and `accomplished_at` anymore

rhnonose: I removed in the other branch, sorry it got messy.

web/models/tasks/task_status.ex
```diff
@@ -0,0 +1,8 @@
+defmodule CoursePlanner.Tasks.TaskStatus do
```
ggpasqualino: this enum is not necessary anymore

web/models/tasks/task.ex
```diff
@@ -0,0 +1,26 @@
...
+    field :deleted_at, :naive_datetime
+
+    timestamps()
+  end
+
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, @cast_params)
+    |> validate_required(@required_params)
+  end
```
ggpasqualino: what do you think of adding also `assoc_constraint` validation for user here?

rhnonose: Yeah, that's a good check, didn't consider it.

PR #46, opened 2017-05-12
Assign courses to terms

ggpasqualino:
- Add OfferedCourse, the association between Term and Course
- Show the courses of a term in the Term show page
- Add/remove courses of term in the Term new/edit pages

### DIFF
--- a/web/models/terms.ex
+++ b/web/models/terms.ex
@@ -2,7 +2,7 @@ defmodule CoursePlanner.Terms do
   @moduledoc """
     Handle all interactions with Terms, create, list, fetch, edit, and delete
   """
-  alias CoursePlanner.Repo
+  alias CoursePlanner.{OfferedCourse, Repo}
   alias CoursePlanner.Terms.Term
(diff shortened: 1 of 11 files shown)


web/models/terms.ex
```diff
@@ -12,17 +12,27 @@ defmodule CoursePlanner.Terms do
...
   def create_term(params) do
     %Term{}
     |> Term.changeset(params)
+    |> Changeset.put_assoc(:offered_courses, course_changesets(params))
     |> Repo.insert
   end
 
+  def course_changesets(%{"course_ids" => ids}) do
+    Enum.map(ids, fn id ->
+      Changeset.cast(%OfferedCourse{}, %{course_id: id}, [:course_id])
```
rhnonose: Isn't it better to put this changeset logic inside of `offered_course.ex`?

ggpasqualino: I gave some thought to it, maybe this is not even the right association to use here, probably it is better to fetch the courses and use the `has_many :courses` association.

I will give it a try

PR #47, opened 2017-05-12
Assign students to courses

ggpasqualino:
- Add CRUD to offered courses (course in a given term)
- Add students to offered courses


ggpasqualino:
I know it is not very nicely decoupled right now, but we haven't moved very fast otherwise

PR #51, opened 2017-05-15
show teacher courses

ghatighorias:


test/controllers/teacher_controller_test.exs
```diff
@@ -66,4 +101,44 @@ defmodule CoursePlanner.TeacherControllerTest do
     conn = get conn, teacher_path(conn, :new)
     assert html_response(conn, 200) =~ "New teacher"
   end
+
+  test "gets empty list when teacher has no course assigned" do
+    teacher = Repo.insert!(%{@user | role: "Teacher"})
```
rhnonose: Can you insert through `Teachers.new/2` with whatever string as second parameter?

test/controllers/teacher_controller_test.exs
```diff
@@ -66,4 +101,44 @@ defmodule CoursePlanner.TeacherControllerTest do
     conn = get conn, teacher_path(conn, :new)
     assert html_response(conn, 200) =~ "New teacher"
   end
+
```
ggpasqualino: those don't test the controller itself, could you move them to `test/lib/course_planner/teachers_test.exs`?

lib/course_planner/teachers.ex
```diff
@@ -48,4 +48,14 @@ defmodule CoursePlanner.Teachers do
   defp format_error({:ok, teacher}, _), do: {:ok, teacher}
   defp format_error({:error, changeset}, teacher), do: {:error, teacher, changeset}
 
+  def courses(teacher_id) do
+    Repo.all(from oc in OfferedCourse,
+      left_join: oct in "offered_courses_teachers", on: oct.offered_course_id == oc.id,
+      left_join: u in User, on: u.id == oct.teacher_id,
```
ggpasqualino: We only use the user id from the user, so we can remove this join and replace the condition
```elixir
where: u.id == ^teacher_id
```
by 
```elixir
where: oct.teacher_id == ^teacher_id
```

lib/course_planner/teachers.ex
```diff
@@ -48,4 +48,14 @@ defmodule CoursePlanner.Teachers do
...
+      join: t in assoc(oc, :term),
+      preload: [term: t],
+      preload: [:course],
```
ggpasqualino: I think you can join the two preloads into
```elixir
preload: [:course, :term]
```

ghatighorias: if I wanna use order_by then I have to assign term: to a variable. then in that list all preloads should be assign to variables and since I wont use the variable for course then i put it in the next line

ghatighorias: preload: [term: t, :course] => this causes error
preload: [term: t, course: c] => excessive assignment to "c"

ggpasqualino: do you need this `t` in the preload? it already binds from `join: t in assoc(oc, :term),`

lib/course_planner/teachers.ex
ghatighorias: I changed but seems to me the result should be the same in both cases

lib/course_planner/teachers.ex
ggpasqualino: okay, I see the point of having the preload like that

if you leave like you did ecto can reuse the query to preload the term, so it does one less query to preload it

👍

lib/course_planner/teachers.ex
ggpasqualino: > so we can remove this join and replace the condition

the whole point here is that we don't need the 
```elixir
left_join: u in User, on: u.id == oct.teacher_id,
```

ghatighorias: that's true. for some reason i was still thinking that user should be joined

PR #60, opened 2017-05-17
Replace course by offered course in the class association

ggpasqualino:


web/templates/class/show.html.eex
```diff
@@ -3,8 +3,13 @@
 <ul>
 
   <li>
+    <strong>Term:</strong>
+    <%= @class.offered_course.term.name %>
+  </li>
+
+  <li>
     <strong>Course:</strong>
-    <%= CourseHelper.get_course_name(@class.course_id) %>
```
rhnonose: Is this function being used anymore? Can you remove it if so?

ggpasqualino: good catch! I will check it

PR #61, opened 2017-05-18
Notify user of updated profile

rhnonose:
Notify user when update action is successful.


lib/course_planner/notifier.ex
```diff
@@ -0,0 +1,33 @@
...
+      {:ok, _} ->
+        {:noreply, state}
+      {:error, reason} ->
+        Logger.error("Email delivery failed: #{reason}")
+        {:noreply, [{:error, reason, email} | state]}
+    end
+  end
+
+  @spec handle_info(any(), any()) :: {:noreply, any()}
+  def handle_info(_, state), do: {:noreply, state}
```
ggpasqualino: why do you need to override `handle_info`?

lib/course_planner/notifier.ex
```diff
@@ -0,0 +1,33 @@
...
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @spec notify_user(User, atom()) :: GenServer.cast
+  def notify_user(%User{} = user, notification_type) do
+    GenServer.cast(__MODULE__, {:send, user, notification_type})
+  end
+
+  @spec handle_cast({atom(), User, atom()}, any()) :: {:noreply, any()}
+  def handle_cast({:send, user, notification_type}, state) do
```
ggpasqualino: we should have one `handle_cast` which consumes the messages that don't match the pattern in order to clean the mailbox

lib/course_planner/notifier.ex
```diff
@@ -0,0 +1,33 @@
...
+    email = UserEmail.build_email(user, notification_type)
+    case Mailer.deliver(email) do
+      {:ok, _} ->
+        {:noreply, state}
+      {:error, reason} ->
+        Logger.error("Email delivery failed: #{reason}")
+        {:noreply, [{:error, reason, email} | state]}
```
ggpasqualino: what's the idea behind saving the error into the GenServer state?

ggpasqualino:
what do you think of sending notification only if the updated user is different from the current user? I think it can be annoying to get notifications about your own modifications

lib/course_planner/notifier.ex
rhnonose: I get spammed with warnings when running `mix test` because the `Swoosh.Adapters.Test` adapter sends a message back to the calling process when e-mail is sent.
I didn't find another good way of solving this issue, but I'm not sure if it's appropriate.

lib/course_planner/notifier.ex
rhnonose: True, I'll add it.

lib/course_planner/notifier.ex
rhnonose: To possibly check not sent e-mails or resend them, either through some buttom or periodically.

rhnonose:
Yeah, I'll add that as well.

lib/course_planner/users.ex
```diff
@@ -33,4 +32,9 @@ defmodule CoursePlanner.Users do
       error -> error
     end
   end
+
+  def notify_user(%{id: id}, %{id: id}, _), do: nil
+  def notify_user(modified_user, _, notification_type) do
```
ggpasqualino: could you rename `_` to `_modifier`, `_current_user` or something which makes sense but with a name?

lib/course_planner/notifier.ex
ggpasqualino: okay, could you add a comment about it?

it is not obvious why we need to override the default implementation of handle_info with a do nothing

ggpasqualino: @rhnonose ping

lib/course_planner/mailer/user_email.ex
```diff
@@ -1,35 +1,23 @@
...
-    Map.update!(email, target_group, &(&1 = [recipient]))
-  end
-
-  @spec create_empty_email(recipient_info) :: Swoosh.Email
-  def create_empty_email(sender_info) do
-    new()
-    |> from(sender_info)
+  def build_email(%{name: _, email: email}, _) when is_nil(email), do: {:error, :invalid_recipient}
+  def build_email(%{name: name, email: email}, notification_type) do
+    case @notifications[notification_type] do
```
ghatighorias: isn't if-else a more straight solution for situation with only 2 cases

web/templates/email/user_updated.html.eex
```diff
@@ -0,0 +1,3 @@
+<div>
+  <h2><%= @name%>, your profile was updated.</h2>
```
ghatighorias: I suggest "your profile was updated successfully"

rhnonose: That would be the case if it's a popup or something, but it's an e-mail notification when someone else updates his profile, so "successfully" doesn't sound right.

lib/course_planner/mailer/user_email.ex
rhnonose: I believe that you should use pattern matching whenever you can in Elixir if it doesn't get too clumped up, and it has better performance (not that it matters too much here).
I also believe I can `%{name: _, email: nil}`

PR #63, opened 2017-05-18
Validate start_date before end_date.

rhnonose:
Validate if start_date is before end_date.
The code looks a bit clunky and I couldn't figure out a better way of doing it, so I'm asking for some feedback @ggpasqualino @ghatighorias .

### DIFF
--- a/web/models/terms/term.ex
+++ b/web/models/terms/term.ex
@@ -5,6 +5,7 @@ defmodule CoursePlanner.Terms.Term do
   use CoursePlanner.Web, :model
 
   alias CoursePlanner.{OfferedCourse, Statuses, Terms.Holiday, Types.EntityStatus}
+  alias Ecto.{Date, Changeset}
 
   schema "terms" do
     field :name, :string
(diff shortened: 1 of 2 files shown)


web/models/terms/term.ex
```diff
@@ -31,5 +32,18 @@ defmodule CoursePlanner.Terms.Term do
     |> validate_required([:name, :start_date, :end_date, :status])
     |> cast_embed(:holidays)
     |> Statuses.update_status_timestamp(EntityStatus)
+    |> validate_changes()
   end
+
+  defp validate_changes(%{data: %{start_date: start_date, end_date: end_date},
```
ggpasqualino: what about naming this `validate_date_range`?

web/models/terms/term.ex
```diff
@@ -31,5 +32,18 @@ defmodule CoursePlanner.Terms.Term do
...
+      changes: changes, errors: []} = changeset) do
+    {:ok, st} = Date.cast(Map.get(changes, :start_date) || start_date)
```
ggpasqualino: If you wanna grab the value regardless of having changes you may use `Ecto.Chanfeset.get_field` instead

rhnonose: I'll test that

web/models/terms/term.ex
```diff
@@ -31,5 +32,17 @@ defmodule CoursePlanner.Terms.Term do
     |> validate_required([:name, :start_date, :end_date, :status])
     |> cast_embed(:holidays)
     |> Statuses.update_status_timestamp(EntityStatus)
+    |> validate_date_range()
   end
+
+  defp validate_date_range(%{errors: []} = changeset) do
```
ggpasqualino: I think it is more usual to check the for `%{valid?: true}` in this case, can you take a look?

web/models/terms/term.ex
```diff
@@ -31,5 +32,17 @@ defmodule CoursePlanner.Terms.Term do
...
+    st = changeset |> Changeset.get_field(:start_date) |> Date.cast!
```
ggpasqualino: do you really need to cast those dates? it seems to me they were supposed to be casted already

rhnonose: I thought so as well, but I was getting some weird things from the difference between native `Date` (`~D` sigil at least) and `Ecto.Date`, since `Term.changeset/2` is used with both types and this was the solution I found to "normalize" the types.

ggpasqualino: aha, weird, but fine

web/models/terms/term.ex
ggpasqualino: @rhnonose ping

rhnonose: sorry, missed this comment

PR #64, opened 2017-05-19
Adds attendance table and module and create functionality

ghatighorias:
create attendance for class students when a class gets created.


lib/course_planner/Factories/factory.ex
```diff
@@ -0,0 +1,56 @@
+defmodule CoursePlanner.Factory do
```
ggpasqualino: what do you think of setting up [ex_machina](https://github.com/thoughtbot/ex_machina) for that?

priv/repo/migrations/20170517092511_create_attendance.exs
```diff
@@ -0,0 +1,16 @@
+defmodule CoursePlanner.Repo.Migrations.CreateAttendance do
+  use Ecto.Migration
+
+  def change do
+    create table(:attendances) do
+      add :attendance_type, :string
```
ggpasqualino: what do you think of using an enum for attendance?

web/controllers/class_controller.ex
```diff
@@ -18,13 +19,20 @@ defmodule CoursePlanner.ClassController do
   def create(conn, %{"class" => class_params}) do
     changeset = Class.changeset(%Class{}, class_params, :create)
 
-    case Repo.insert(changeset) do
+    multi =
+      Multi.new
+      |> Multi.insert(:class, changeset)
```
ggpasqualino: We should have one transaction to create all attendances together, but I am not sure about the class and the attendances being created in the same transaction

What can the user do if the transaction fails because of the attendances?

lib/course_planner/Factories/factory.ex
ghatighorias: hmm nice, didn't know that one

priv/repo/migrations/20170517092511_create_attendance.exs
ghatighorias: so this is a tricky question.
cause we can put this option to setting so client can change it, or we can hardcode them

web/controllers/class_controller.ex
ghatighorias: so my understanding was the transaction runs when someone triggers "Repo.transaction(multi)" . and multi has the list of all changes required.

ggpasqualino: yes, so?

web/models/class/class.ex
```diff
@@ -31,6 +32,13 @@ defmodule CoursePlanner.Class do
     |> validate_required([:offered_course_id, :date, :starting_at, :finishes_at, :status])
   end
 
+  def changeset(struct, params, :fill) do
+    struct
+    |> cast(params, [:course_id])
```
ggpasqualino: why do you need the `course_id` when you fill the class attendance?

web/models/class/class_helper.ex
```diff
@@ -38,6 +38,46 @@ defmodule CoursePlanner.ClassHelper do
...
 
   defp non_deleted_query do
-    from c in Class , where: is_nil(c.deleted_at)
+    from c in Class,
+      preload: [{:offered_course, :course}],
+      where: is_nil(c.deleted_at),
+      order_by: [desc: :date, desc: :starting_at]
+  end
+
+  defp get_class_students(offered_course_id) do
```
ggpasqualino: what do you think of adding `has_many :students, through: [:offered_course, :students]` in the class schema?

this way you don't need to add those joins every time, only the preload

web/controllers/class_controller.ex
ghatighorias: ahhhh. now i see what you mean. pfffff such a mistake

web/models/class/class_helper.ex
rhnonose: 👍

priv/repo/migrations/20170517092511_create_attendance.exs
ggpasqualino: okay, let's leave it like that for now then

web/controllers/class_controller.ex
ghatighorias: Tests show that it works fine since in Multi.run

web/models/class/class.ex
ghatighorias: this i wrote before you change to offered_course_id. I have to change it to "offered_course_id"

web/models/class/class_helper.ex
ghatighorias: is this good now ?

test/controllers/class_controller_test.exs
```diff
@@ -253,4 +253,33 @@ defmodule CoursePlanner.ClassControllerTest do
...
+
+  test "creates class and all attendance", %{conn: conn} do
+    course = insert(:course)
+    term1 = insert(:term, %{
+                            start_date: %Ecto.Date{day: 1, month: 1, year: 2017},
+                            end_date: %Ecto.Date{day: 1, month: 6, year: 2017},
+                            courses: [course]
+                           })
+
+    students =
```
ggpasqualino: you may use the `insert_list(3, :user, %{role: "Student"})` instead

test/controllers/class_controller_test.exs
```diff
@@ -253,4 +253,33 @@ defmodule CoursePlanner.ClassControllerTest do
...
+      [
+         insert(:user, %{role: "Student"}),
```
ggpasqualino: what do you think of creating a `student_factory` in the factories file that already adds the role?

test/controllers/class_controller_test.exs
```diff
@@ -253,4 +253,28 @@ defmodule CoursePlanner.ClassControllerTest do
...
+    class_attrs = %{@valid_attrs | offered_course_id: offered_course.id, status: "Active"}
+
+    conn = post conn, class_path(conn, :create), class: class_attrs
+    assert redirected_to(conn) == class_path(conn, :index)
+    assert Repo.get_by(Class, class_attrs)
+
+    class = List.first(ClassHelper.all_none_deleted())
+    attendances = AttendanceHelper.get_class_attendance_info(class.id)
+
+    Enum.map(students, fn(item_student) ->
```
ggpasqualino: Can you explain to me what you want to assert here? this is confusing to me

ghatighorias: I am checking that the students that are having attendance are the exact ones attendance been created for them

web/models/class/class_helper.ex
```diff
@@ -38,6 +39,42 @@ defmodule CoursePlanner.ClassHelper do
...
+    Repo.all(
+    from c in Class,
+      join: oc in assoc(c, :offered_course),
+      preload: [:students, offered_course: oc],
+      where: oc.id == ^offered_course_id
+      )
+  end
+
+  def create_class_attendance_records(class) do
+    class_data = List.first(get_class_students(class.offered_course_id))
```
ggpasqualino: now you can access the students of class with `class.students` if you preload the students association

are you sure you even need to call this method?

test/controllers/class_controller_test.exs
ggpasqualino: maybe we can make this assertion less smart but more understandable

ggpasqualino: what do you think of the following test instead?
```elixir
  test "creates class and all attendance", %{conn: conn} do
    course = insert(:course)
    term1 = insert(:term, %{
                            start_date: %Ecto.Date{day: 1, month: 1, year: 2017},
                            end_date: %Ecto.Date{day: 1, month: 6, year: 2017},
                            courses: [course]
                           })

    students = insert_list(3, :student)
    offered_course = insert(:offered_course, %{term: term1, course: course, students: students})
    class_attrs = %{@valid_attrs | offered_course_id: offered_course.id, status: "Active"}

    conn = post conn, class_path(conn, :create), class: class_attrs
    assert redirected_to(conn) == class_path(conn, :index)
    assert class = Repo.get_by(Class, class_attrs) |> Repo.preload(:attendances)

    student_ids = students |> Enum.map(&(&1.id)) |> Enum.sort()
    attendance_student_ids = class.attendances |> Enum.map(&(&1.student_id)) |> Enum.sort()

    assert student_ids == attendance_student_ids
  end
```

web/models/attendance/attendance_helper.ex
```diff
@@ -0,0 +1,15 @@
+defmodule CoursePlanner.AttendanceHelper do
```
ggpasqualino: can we remove this module? it is not used anywhere anymore

ggpasqualino: @ghatighorias ping

PR #66, opened 2017-05-22
Notify user when term has changed

rhnonose:
Will be left out the notification when creating a new term since there's no user connected when a term is created.


web/models/terms.ex
```diff
@@ -70,4 +70,19 @@ defmodule CoursePlanner.Terms do
   defp non_deleted_query do
     from t in Term, where: is_nil(t.deleted_at)
   end
+
+  def notify_term_users(term, current_user, notification_type) do
+    term
+    |> get_enrolled_students()
```
ghatighorias: I think literally everyone except volunteers needs to be informed about the changes in the term

ghatighorias: so you can inform teachers of the term and organizers

web/models/terms.ex
```diff
@@ -70,4 +70,19 @@ defmodule CoursePlanner.Terms do
...
+    |> Enum.reject(fn %{id: id} -> id == current_user.id end)
+    |> Enum.each(&(Notifier.notify_user(&1, notification_type)))
+  end
+
+  defp get_enrolled_students(term) do
+    term
+    |> Repo.preload([:offered_courses, offered_courses: :students])
+    |> Map.get(:offered_courses)
+    |> Enum.map(&(Map.get(&1, :students)))
+    |> List.flatten()
```
ggpasqualino: what do you think about using `Enum.flat_map/2` instead?

web/models/terms.ex
```diff
@@ -70,4 +70,19 @@ defmodule CoursePlanner.Terms do
...
+    |> get_enrolled_students()
+    |> Enum.reject(fn %{id: id} -> id == current_user.id end)
+    |> Enum.each(&(Notifier.notify_user(&1, notification_type)))
+  end
+
+  defp get_enrolled_students(term) do
+    term
+    |> Repo.preload([:offered_courses, offered_courses: :students])
+    |> Map.get(:offered_courses)
+    |> Enum.map(&(Map.get(&1, :students)))
```
ggpasqualino: Sorry, maybe I was not very clear, what I meant is to replace
```elixir
|> Enum.map(&(Map.get(&1, :students)))
|> List.flatten()
```
by
```elixir
|> Enum.flat_map(&(Map.get(&1, :students)))
```

PR #67, opened 2017-05-22
Notify user when a course has changed

rhnonose:
Will be left out the notification when creating a new course since there's no user connected when a course is created.


lib/course_planner/mailer/user_email.ex
```diff
@@ -5,7 +5,9 @@ defmodule CoursePlanner.Mailer.UserEmail do
   use Phoenix.Swoosh, view: CoursePlanner.EmailView, layout: {CoursePlanner.LayoutView, :email}
 
   @notifications %{
-    user_modified: %{subject: "Your profile was updated", template: "user_updated.html"}
+    user_modified: %{subject: "Your profile was updated", template: "user_updated.html"},
```
ghatighorias: is updated

lib/course_planner/mailer/user_email.ex
```diff
@@ -5,7 +5,9 @@ defmodule CoursePlanner.Mailer.UserEmail do
...
+    course_updated: %{subject: "A course you subscribe to was updated", template: "course_updated.html"},
+    course_deleted: %{subject: "A course you subscribe to was deleted", template: "course_deleted.html"},
```
ghatighorias: subscribed to

lib/course_planner/mailer/user_email.ex
rhnonose: Since it's a notification that someone else updated it, should be "was".

web/models/course/course_helper.ex
```diff
@@ -28,4 +27,29 @@ defmodule CoursePlanner.CourseHelper do
...
+  end
+
+  defp get_subscribed_users(course) do
+    offered_courses = course
+    |> Repo.preload([:offered_courses, offered_courses: :students, offered_courses: :teachers])
+    |> Map.get(:offered_courses)
+
+    students = offered_courses
+    |> Enum.map(&(Map.get(&1, :students)))
+    |> List.flatten()
```
ggpasqualino: what do you think about using `Enum.flat_map/2` instead?

lib/course_planner/mailer/user_email.ex
ghatighorias: nope, since it's a passive of simple present it should be "is updated".
cause the event happened just now and not somewhere in the past

web/models/course/course_helper.ex
ggpasqualino: ping

PR #69, opened 2017-05-26
Deletes all attendances when a class is hard deleted

ghatighorias:


test/controllers/class_controller_test.exs
```diff
@@ -275,4 +275,53 @@ defmodule CoursePlanner.ClassControllerTest do
...
+    term1 = insert(:term, %{
+                            start_date: %Ecto.Date{day: 1, month: 1, year: 2017},
+                            end_date: %Ecto.Date{day: 1, month: 6, year: 2017},
+                            courses: [course]
+                           })
+
+    students = insert_list(3, :student)
+    offered_course = insert(:offered_course, %{term: term1, course: course, students: students})
+    class_attrs = %{@valid_attrs | offered_course_id: offered_course.id, status: "Planned"}
+    class = insert(:class, class_attrs)
```
ggpasqualino: As I seek, the important parts here are the class, students and attendances.

I think we can define associations on the factories so we don't need to create everything here

web/templates/attendance/index.html.eex
```diff
@@ -0,0 +1,23 @@
...
+      <th>Teachers</th>
+      <th></th>
+    </tr>
+  </thead>
+  <tbody>
+    <%= for offered_course <- @offered_courses do %>
+        <tr>
+          <td><%= offered_course.term.name %></td>
+          <td><%= offered_course.course.name %></td>
+          <td><%= Enum.reduce(offered_course.teachers, "", fn(teacher, out) ->
```
ggpasqualino: what do you think of extracting this whole logic
```elixir
Enum.reduce(offered_course.teachers, "", fn(teacher, out) ->
  Enum.join([out, teacher.name, teacher.family_name], " ")
end)
```
into a function in the `AttendanceView` module?

web/router.ex
```diff
@@ -47,6 +47,8 @@ defmodule CoursePlanner.Router do
     resources "/tasks", TaskController do
       post "/grab", TaskController, :grab, as: :grab
     end
+
+    resources "/attendance", AttendanceController, only: [:index]
```
ggpasqualino: the path of the resource `"/attendance"` is usually plural, could you change it please?

web/models/attendance/attendance_helper.ex
```diff
@@ -0,0 +1,37 @@
...
+    Repo.all(from oc in OfferedCourse,
+      join: t in assoc(oc, :teachers),
+      preload: [:term, :course, teachers: t],
+      where: t.id == ^teacher_id)
+  end
+
+  def get_all_student_offered_courses(student_id) do
+    Repo.all(from oc in OfferedCourse,
+      join: s in assoc(oc, :students),
+      preload: [:term, :course, students: s],
```
ggpasqualino: you always show the offered course's teachers in the index template, so it would be good to preload `:teachers` here as well

web/models/attendance/attendance_helper.ex
```diff
@@ -0,0 +1,37 @@
+defmodule CoursePlanner.AttendanceHelper do
+  @moduledoc """
+  This module provides custom functionality for controller over the model
+  """
+  use CoursePlanner.Web, :model
+
+  alias CoursePlanner.{Repo, OfferedCourse}
+
+  def get_course_attendances(offered_course_id) do
```
ggpasqualino: this function doesn't seem to be used anywhere, can we remove it?

web/controllers/attendance_controller.ex
```diff
@@ -0,0 +1,23 @@
+defmodule CoursePlanner.AttendanceController do
```
ggpasqualino: why do you pass `role: :coordinator`, `role: :teacher` and `role: :student` to the render function?

ggpasqualino:
could you split the index part into a new pr please?

web/controllers/attendance_controller.ex
ghatighorias: I was thinking of in the next pull request I can filter "fill" or "show" functionality based on the role.

web/models/attendance/attendance_helper.ex
ghatighorias: will be used in the next pull requests. I forgot to remove it from this one

PR #70, opened 2017-05-29
Send resource link in e-mail notifications

rhnonose:
To build the resource link in the template, we would need `Plug.Conn` passed around which I thought it was bad idea, so I'm sending the path through a parameter.
It might be getting a little bit clunky with too many parameters, but I couldn't think of a better way, so please take a look @ggpasqualino @ghatighorias if you have a better idea.

### DIFF
--- a/lib/course_planner/notifier.ex
+++ b/lib/course_planner/notifier.ex
@@ -33,13 +33,13 @@ defmodule CoursePlanner.Notifier do
   end
 
   @spec handle_cast({atom(), Notification.t}, any()) :: {:noreply, any()}
-  def handle_cast({:send_email, %{to: user, type: type, resource_path: path}}, state) do
-    email = UserEmail.build_email(user, type, path)
+  def handle_cast({:send_email, notification}, state) do
+    email = UserEmail.build_email(notification)
     case Mailer.deliver(email) do
       {:ok, _} ->
         {:noreply, state}
       {:error, reason} ->
-        Logger.error("Email delivery failed: #{reason}")
+        Logger.error("Email delivery failed: #{Kernel.inspect reason}")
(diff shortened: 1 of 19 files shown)


ggpasqualino:
Regarding passing the path from the controller I think that's a good solution

What do you think of creating a Notification structure which contains this `path` and the `notification_type`? With this in place we don't need to change all the function signatures when we need to add more things to the notification

rhnonose:
Sounds good.

rhnonose:
I added a nested module for the struct, but I'm still wrapping my head around this type of design.
One thing I might have done is to have the functions defined in the outer module and that would build the `Notification` struct without having to have it imported by the caller, but still not sure as well, so I would like some feedback.

lib/course_planner/notifier.ex
```diff
@@ -6,19 +6,35 @@ defmodule CoursePlanner.Notifier do
...
+  def notify_user(%Notification{} = notification) do
+    GenServer.cast(__MODULE__, {:send_email, notification})
   end
 
-  @spec handle_cast({atom(), User, atom()}, any()) :: {:noreply, any()}
-  def handle_cast({:send, user, notification_type}, state) do
-    email = UserEmail.build_email(user, notification_type)
+  @spec handle_cast({atom(), Notification.t}, any()) :: {:noreply, any()}
+  def handle_cast({:send_email, %{to: user, type: type, resource_path: path}}, state) do
+    email = UserEmail.build_email(user, type, path)
```
ggpasqualino: what do you think of passing the notification down instead of destructuring it?

rhnonose: I left it that way to change less function signatures before feedback, so yeah we can do that as well, it would get cleaner.

PR #71, opened 2017-05-29
Replace with to update association correctly.

rhnonose:


ggpasqualino:
@rhnonose can you comment why you closed it?

rhnonose:
@ggpasqualino The fix seems correct but my regression test was wrong and it's breaking another test, I'm figuring it out.

PR #74, opened 2017-05-30
Remove course manipulation from term scope

rhnonose:
Remove course manipulation from terms as we discussed since `OfferedCourse` already provides the needed functionality.

### DIFF
--- a/web/models/terms.ex
+++ b/web/models/terms.ex
@@ -24,6 +24,7 @@ defmodule CoursePlanner.Terms do
   def get(id) do
     non_deleted_query()
     |> Repo.get(id)
+    |> Repo.preload([:courses])
   end
 
   def edit(id) do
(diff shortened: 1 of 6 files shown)


web/models/terms.ex
```diff
@@ -18,19 +18,12 @@ defmodule CoursePlanner.Terms do
...
 
-  def course_changesets(%{"course_ids" => ids}) do
-    Enum.map(ids, &OfferedCourse.add_to_term_changeset/1)
-  end
-  def course_changesets(_), do: []
-
   def get(id) do
     non_deleted_query()
     |> Repo.get(id)
-    |> Repo.preload(:courses)
```
ggpasqualino: maybe showing the courses still makes sense

web/templates/term/show.html.eex
```diff
@@ -37,20 +37,6 @@
       <% end %>
     </table>
   <% end %>
-
```
ggpasqualino: maybe showing the courses still makes sense

ggpasqualino:
I think we can remove also `OfferedCourse.add_to_term_changeset/1`

PR #76, opened 2017-05-31
Add restriction to Term

ggpasqualino:
Students, Teachers, and Volunteers can't do anything with Terms

### DIFF
--- a/lib/course_planner/permissions.ex
+++ b/lib/course_planner/permissions.ex
@@ -1,13 +1,19 @@
 defimpl Canada.Can, for: CoursePlanner.User do
-  alias CoursePlanner.User
-  alias CoursePlanner.Tasks.Task
+  alias CoursePlanner.{Tasks.Task, Terms.Term, User}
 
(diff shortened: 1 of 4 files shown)


lib/course_planner/permissions.ex
```diff
@@ -1,13 +1,19 @@
 defimpl Canada.Can, for: CoursePlanner.User do
-  alias CoursePlanner.User
-  alias CoursePlanner.Tasks.Task
+  alias CoursePlanner.{Tasks.Task, Terms.Term, User}
 
   def can?(%User{role: "Coordinator"}, _, _), do: true
+
+  def can?(%User{role: role}, _action, Term)
+    when role in ["Teacher", "Student", "Volunteer"], do: false
+  def can?(%User{role: role}, _action, %Term{})
```
ghatighorias: So "Teacher", "Student", "Volunteer" won't be able to see the term ?

ggpasqualino: yeah, the CRUD pages at least we set in the workflow that they shouldn't see it

PR #77, opened 2017-05-31
Laszlo.kardinal/login screen

laszlokardinal-dina:


web/static/js/app.js
```diff
@@ -1,59 +1,11 @@
...
-// Local files can be imported directly using relative
-// paths "./socket" or full ones "web/static/js/socket".
+  require( "material-design-lite" );
 
-// import socket from "./socket"
+  /*
+      APPLICATION
+  */
 
-window.onload = () => {
```
ggpasqualino: my bad, I told you could remove all the existing stuff, but this is being used in the terms new/edit page for adding/removing holidays

can you add this back? or maybe reimplement better than I did 😬

web/static/js/app.js
```diff
@@ -1,59 +1,13 @@
...
-// However, those files will only be executed if
-// explicitly imported. The only exception are files
-// in vendor, which are never wrapped in imports and
-// therefore are always executed.
 
-// Import dependencies
-//
-// If you no longer want to use a dependency, remember
-// to also remove its path from "config.paths.watched".
-import "phoenix_html"
```
ggpasqualino: The delete links broke, maybe we should add this back 🤔

PR #79, opened 2017-06-02
Adds log out functionality to the layout

ghatighorias:


ghatighorias:
I am not sure exactly what is the best way of showing the user and is log-out link. tell if there is any idea

web/controllers/coherence/session_controller.ex
```diff
@@ -0,0 +1,323 @@
+defmodule CoursePlanner.Coherence.SessionController do
```
ggpasqualino: we shouldn't add this file

I checked it out and the reason you need to add it is that we have the coherence protected routes wrong.

Could you move
```elixir
coherence_routes :protected
```
into it's own block?
```elixir
  scope "/" do
    pipe_through :protected
    coherence_routes :protected
  end
```
then remove this file?

ggpasqualino:
> I am not sure exactly what is the best way of showing the user and is log-out link. tell if there is any idea

@ghatighorias that's fine for now, later there will be lateral navigation bar and a header with user info and logout link

web/controllers/coherence/session_controller.ex
ghatighorias: cool, 
adding the session_controller was the only option i have found but it was pretty ugly

PR #81, opened 2017-06-06
Adds permission for class controller

ghatighorias:


web/controllers/class_controller.ex
```diff
@@ -39,11 +42,16 @@ defmodule CoursePlanner.ClassController do
...
-    render(conn, "show.html", class: class)
+    case Repo.get(Class, id) do
+      nil ->
+        conn
+        |> put_status(404)
+        |> render(CoursePlanner.ErrorView, "404.html")
+      class ->
+        class
+        |> Repo.preload([:offered_course, offered_course: :term, offered_course: :course])
+        |> (&render(conn, "show.html", class: &1)).()
```
ggpasqualino: I think this could be more understandable without the pipe and the anonymous function

what do you think of the following?
```elixir
class = Repo.preload(class, [:offered_course, offered_course: :term, offered_course: :course])
render(conn, "show.html", class: class)
```

PR #85, opened 2017-06-06
Replace resource path by url

ggpasqualino:
Before we were sending notifications with broken links because they had
only the path of the resource instead of the full url.


rhnonose:
😅

PR #88, opened 2017-06-06
Fixes bug that allowed a course be offered multiple times in a term

ghatighorias:
* With this validation each course can be offered only once in each term.


rhnonose:
Isn't it better to have a composite primary key involving `course_id` and `term_id`?

ghatighorias:
@rhnonose honestly i didn't know composite keys are possible. @ggpasqualino what is your opinion ? shall I change it according to what @rhnonose said ?

ggpasqualino:
I agree with @rhnonose in the sense that we should let this validation to the database, but I wouldn't replace the current id by a composite primary key, they are annoying to work with, specially with all the associations which would require this composite key.

My suggestion is to have an unique index with both fields and then you can add unique_constraint into the changeset.

you can check more about it https://hexdocs.pm/ecto/Ecto.Changeset.html#unique_constraint/3

PR #91, opened 2017-06-08
Redirect root route to dashboard

ggpasqualino:


ggpasqualino:
@rhnonose no, that's what I was explaining in the standup, we need a redirect because the "/" as a  whole path messes the design

ggpasqualino:
@rhnonose another idea though is to remove the page controller and add a redirector plug 🤔

rhnonose:
Got it. Not a bad idea, your call.

PR #93, opened 2017-06-08
Introduces Setting page for the courseplanner

ghatighorias:


web/models/setting/setting.ex
```diff
@@ -0,0 +1,37 @@
+defmodule CoursePlanner.Setting do
+  @moduledoc """
+  This module holds the model for the setting table
+  """
+  use CoursePlanner.Web, :model
+
+  schema "settings" do
+    field :notification_frequency, :integer
```
rhnonose: What's the meaning of this field? Nowhere it's clear what it means or does.

ghatighorias: maybe the naming is bad. but It is suppose to be used for how often attendance notification should be send to teachers. the number suppose to be in days. (every 1 day, every 2 day and so on)

PS : I can additionally add a boolean for it to be sent at all or not

but if you think the naming is not good do tell, cause I had my mind froze on this

rhnonose: I got it, maybe you can add a comment in `@moduledoc` or even a tooltip in the view explaining the functionality.

ggpasqualino: this is rather the notification period, frequency is more like 3 notifications/day

priv/repo/migrations/20170607142123_create_setting.exs
```diff
@@ -0,0 +1,17 @@
+defmodule CoursePlanner.Repo.Migrations.CreateSetting do
+  use Ecto.Migration
+
+  def change do
+    create table(:settings) do
```
ggpasqualino: I was thinking about this table...

What do you @ghatighorias @rhnonose  think about having the settings table with the columns name and value and each setting is a different row? Somewhat like a key-value store

ggpasqualino: The advantage is that it would be easier to add and remove values because they are not coupled to the schema

The disadvantage is that the ui needs to be dynamic to add new fields, and maybe we can't validate the field type

ghatighorias: hmm, I see and I should say it's a very good idea. Although it makes more job rather for frontend :P
But I generally agree. @rhnonose what is your idea ?

PS : do I get this correctly 

1. Keys should be unique in the database and application level
2. in the seed we will add all the setting we need already

ggpasqualino: yeah, that's the idea

rhnonose: I think it's a good idea and maybe there's a way of getting the benefit of both ways.
There's a way of using ecto schemas but not necessarily persisting it in a direct mapping, meaning: we can have a schema like the current with parameter fields, types and validation but persist in a key-value way. Database validations are lost but removes the necessity of migration.

ghatighorias: hmm can you give me a link @rhnonose

rhnonose: I don't remember where I read it but it wasn't a concrete example.
You can start by checking the `:map` type in ecto.

ghatighorias: Done

web/templates/setting/form.html.eex
```diff
@@ -0,0 +1,28 @@
...
+        <th>Name</th>
+        <th>Value</th>
+      </tr>
+    </thead>
+    <tbody>
+      <%= for svc <- @system_variable_changesets do %>
+        <tr>
+          <td> <%= svc.data.key %> </td>
+
+          <%= inputs_for f, integer_to_atom(svc.data.id), fn ff -> %>
```
rhnonose: This looks like a bad idea. I can keep creating settings and overflow the atoms table.

priv/repo/seeds.exs
```diff
@@ -9,3 +10,39 @@ CoursePlanner.User.changeset(%CoursePlanner.User{},
     role: "Coordinator"},
     :seed)
 |> CoursePlanner.Repo.insert!
+
+CoursePlanner.SystemVariable.changeset(%CoursePlanner.SystemVariable{},
+  %{
+    key: "Program name",
```
ggpasqualino: I would suggest the keys to have a more strict letter case, like screaming snake case `PROGRAM_NAME`

test/controllers/setting_controller_test.exs
```diff
@@ -0,0 +1,122 @@
...
+  end
+
+  test "shows chosen resource only for coordinator", %{conn: _conn} do
+    coordinator_conn = login_as(:coordinator)
+
+    conn = get coordinator_conn, setting_path(coordinator_conn, :show)
+    assert html_response(conn, 200) =~ "Show settings"
+  end
+
+  test "does not show chosen resource for non coordinator users", %{conn: _conn} do
```
ggpasqualino: Could you check something like the offered_course_controller_test.exs? 

These tests contain too much information, and in the other hand if they fail there is not enough information, so it is hard to understand why they failed

test/models/system_variable_test.exs
```diff
@@ -0,0 +1,118 @@
+defmodule CoursePlanner.SystemVariableTest do
```
ggpasqualino: please break these tests apart, if you want to group them because they are kinda related you can use the `describe`

web/models/setting/settings.ex
```diff
@@ -0,0 +1,79 @@
...
+      Enum.reduce(changesets, Multi.new, fn(changeset, out_multi) ->
+
+        case changeset do
+           :non_existing_resource -> Multi.error(out_multi, :non_existing_resource, "Resource does not exist")
+           :uneditable_resource   -> Multi.error(out_multi, :uneditable_resource, "Resource is not editable")
+           _  ->
+             operation_atom =
+               changeset.data.id
+               |> Integer.to_string()
+               |> String.to_atom()
```
ggpasqualino: I don't think you need an atom here, and as @rhnonose mentioned this can be a bad idea to convert everything to atom

web/models/setting/settings.ex
```diff
@@ -0,0 +1,79 @@
...
+  def insert_error(form, atomized_group_id, errors) do
+    releavant_error =
+      errors
+      |> Enum.find(fn(error) -> error.failed_id == atomized_group_id end)
+
+    if is_nil(releavant_error) do
+      form
+    else
+      form = Map.put form, :valid?, false
+      add_error(form, :value, releavant_error.message)
```
ggpasqualino: doesn't `add_error` put the `valid?: false` already?

web/models/setting/system_variable.ex
```diff
@@ -0,0 +1,96 @@
...
+
+    struct
+    |> cast(params, target_params)
+    |> validate_required(target_params)
+    |> validate_value_type()
+  end
+
+  def changeset(struct, params, :update) do
+    struct
+    |> changeset(params)
```
ggpasqualino: How do you intend to use a system variable?

Let's suppose you want to do something like the following
```elixir
if SystemVariable.get("SEND_NOTIFICATION_ATTENDANCE", default: false) do
  ...
end
```

then you should not cast the value for `key` because this will break the system

web/models/setting/system_variable.ex
```diff
@@ -0,0 +1,96 @@
...
+    |> validate_editable()
+  end
+
+  defp validate_editable(%{valid?: true} = changeset) do
```
ggpasqualino: for this to work properly you should not cast the value for `editable` in the update changeset, otherwise it can be bypassed by changing the editable field as well

web/models/setting/system_variable.ex
```diff
@@ -0,0 +1,96 @@
...
+    case validation_result do
+      {:ok, _} -> changeset
+      {:error, message} -> add_error(changeset, :value, message)
+    end
+  end
+  defp validate_value_type(changeset), do: changeset
+
+  def parse_integer(value) do
+    case Integer.parse(value) do
+      {parsed_value, ""} -> validate_number_rage_inclusive(parsed_value, 0, 1_000_000)
```
ggpasqualino: elixir has ranges in the std library, you can do the same check as
```elixir
parsed_value in (0..1_000_000)
```

web/models/setting/settings.ex
```diff
@@ -0,0 +1,79 @@
...
+
+  def get_changesets_for_update(setting_params) do
+    system_variables = all()
+    setting_params = Enum.sort_by(setting_params, &(elem(&1, 1)["key"]))
+
+    Enum.map(setting_params, fn(setting_param) ->
+        {param_id, %{"key" => _param_key, "value" => param_value}} = setting_param
+
+      found_system_variable =
+        Enum.find(system_variables, fn(system_variable) ->
```
ggpasqualino: what do you think of using a map with the ids as keys instead of a list?

web/templates/setting/form.html.eex
ghatighorias: No you cannot do that cause the user can just have edit and update functionality. then in the controller I read the existing settings and use the ID to organize the changes of the settings. So I don't see how the user can create settings at all

web/models/setting/settings.ex
ghatighorias: since in the view I use Atom for the fields then the operation ID which I use here have to be of the same type to show the error correctly

web/models/setting/system_variable.ex
ghatighorias: But "validate_number_rage_inclusive" adds the error correctly itself. converting to your solution then I have to include an add_error instead of validate_number_rage_inclusive which to me both are the same

web/models/setting/settings.ex
ghatighorias: I am not sure if I get your point correctly. But in case you are talking about "setting_params" then they are coming from the form or if it is about "system_variables" then it's directly from database

web/templates/setting/form.html.eex
ghatighorias: Or in case I cannot get the problematic part please do explain more

web/models/setting/system_variable.ex
ghatighorias: done

web/models/setting/system_variable.ex
ghatighorias: done

test/controllers/setting_controller_test.exs
ghatighorias: I am new to 'describe' but I tried to do my best

test/models/system_variable_test.exs
ghatighorias: I am new to 'describe' but I tried to do my best

test/controllers/setting_controller_test.exs
ghatighorias: do you mean to break all this test ? I am okay if you don't think they gonna be much

ghatighorias: done

web/templates/setting/form.html.eex
rhnonose: I just saw that the user can't create params. But it's still slightly problematic because it's coming from the database and since it's supposed to be a stand-alone app, we can't make sure the table doesn't grow.

test/controllers/setting_controller_test.exs
ggpasqualino: > do you mean to break all this test ? I am okay if you don't think they gonna be much

yeah, I would rather have more tests which are smaller and straight forward

web/templates/setting/form.html.eex
ghatighorias: Seems I fail to see the issue.
whenever you make a form the data is formulated like what I did. in fact I replicated what form generates when it has a valid changeset. what it basically does is creating a map of .   %{id: %{key: key, value: value}}

Anyhow, how to do you suggest me change this ?

rhnonose: Yes, so suppose you have a table with 10000 entries (which through the interface you can't create but you can do that accessing the database directly, which we can't guarantee it's not done). Each id is converted to an atom and placed in the atom's table. This may overflow the memory space allocated to the atom table which has a fixed amount and is not garbage collected.
I'm thinking about a solution, but can't find anything related in the form to replace the usage of atom for the fields. We can set an arbitrary number to limit the number of rows, but that's a monkey patch approach.

ghatighorias: Ok I didn't know that. I read this article http://blog.pemedeiros.com/elixir-atoms-are-everywhere/ which says "The Erlang vm stores the Atom inside a table internally and has a default limit of 1048576 atoms per instance. This value can be configured, but in the end it is not an optimization that we want to make."
Honestly it's a very stupid limitation, but in any case I am still thinking about other forms that are using atoms. So in any form that one uses inputs_for, data gets sub-grouped by atoms which is referred to rows.

I start checking my solution to see how else I can change it

ghatighorias: Another issue is that the "field" section in Inputs_for just accept atoms

web/models/setting/system_variable.ex
ggpasqualino: sorry, I meant the implementation of the function `validate_number_rage_inclusive`

web/models/setting/settings.ex

web/models/setting/system_variable.ex
ghatighorias: done

web/controllers/setting_controller.ex
```diff
@@ -0,0 +1,43 @@
...
+  plug :authorize_controller
+
+  def show(%{assigns: %{current_user: %{role: "Coordinator"}}} = conn, _param) do
+    system_variables = Settings.get_visible_systemvariables()
+    render(conn, "show.html", system_variables: system_variables)
+  end
+
+  def edit(%{assigns: %{current_user: %{role: "Coordinator"}}} = conn, _param) do
+    variables = Settings.get_editable_systemvariables()
+    changesets = Enum.map(variables, fn(variable) -> SystemVariable.changeset(variable) end)
```
ggpasqualino: when you use a function with arity 1 in the map you can use it as following
```
Enum.map(variables, &SystemVariable.changeset/1)
```

web/models/setting/settings.ex
```diff
@@ -0,0 +1,75 @@
...
+    Repo.all(@visible_settings_query)
+  end
+
+  def get_editable_systemvariables do
+    Repo.all(@editable_settings_query)
+  end
+
+  def get_changesets_for_update(setting_params) do
+    system_variables = all() |> Enum.into(Map.new)
+    setting_params = Enum.sort_by(setting_params, &(elem(&1, 1)["key"]))
```
ggpasqualino: why is it necessary to sort those?

web/models/setting/system_variable.ex
```diff
@@ -0,0 +1,104 @@
...
+      _                  -> {:error, "the given value should be a number"}
+    end
+  end
+
+  defp validate_number_rage_inclusive(number, min, max) do
```
ggpasqualino: validate_number_***rage***_inclusive 😡  typo? :D

web/templates/setting/form.html.eex
rhnonose: If there's no immediate solution, we can leave it at that and later implement our own `inputs_for` to remove the atoms limitation. This PR is big enough that you can sort this detail out of it.

web/models/setting/settings.ex
ghatighorias: Since we are iterating based on the setting that are coming from form it's not for sure that they will come in a specific order, so I have to sort them like I order the read so Show, edit and update render the system_variables in the same way.

web/models/setting/settings.ex
```diff
@@ -0,0 +1,75 @@
...
+             Multi.update(out_multi, operation_atom, changeset)
+        end
+      end)
+
+    Repo.transaction(multi)
+  end
+
+  def insert_error(form, atomized_group_id, errors) do
+    releavant_error =
+      errors
```
ggpasqualino: can we make `errors` a map which keys are the `failed_id` instead of a list?

web/models/setting/system_variable.ex
ghatighorias: :D :D sorry

ghatighorias: done

web/controllers/setting_controller.ex
ghatighorias: done

web/templates/setting/form.html.eex
ghatighorias: in case you are interested the problem is in "to_form" function in  phoenix

ggpasqualino:
@ghatighorias @rhnonose  I made some changes to make this closer to the standard way of building forms, please take a look

rhnonose:
LGTM

ghatighorias:
I am looking to see why all the tests are failing with "cannot convert struct to param, key :id contains a nil value"

ggpasqualino:
@ghatighorias you did some weird merge of master into this branch
image

ghatighorias:
I think this was a github mistake, it told me to update the branch and that it's successful and then it says 0 lines changed

ggpasqualino:
@ghatighorias I don't know, I reverted your merge commit and it doesn't ask anything like that

anyways, could you please check the refactoring I did?

ghatighorias:
ok sorry my bad, my database was dropped :P
@ggpasqualino I am on it

lib/course_planner/permissions.ex
```diff
@@ -39,6 +41,9 @@ defimpl Canada.Can, for: CoursePlanner.User do
   def can?(%User{role: role}, _action, %User{})
     when role in ["Teacher", "Student", "Volunteer"], do: false
 
+  def can?(%User{role: "Coordinator"}, _action, SettingController), do: true
```
ghatighorias: Coordinator is already has the right in the following line "def can?(%User{role: "Coordinator"}, _, _), do: true"

ggpasqualino: okaaay, but that's your own code 😂

ghatighorias: yes :P hahahha
I change and commit it now

ggpasqualino:
fyi my changes are the following https://github.com/digitalnatives/course_planner/pull/93/commits/0a799ff987fa11e819758f8e74744a3fc3175385 https://github.com/digitalnatives/course_planner/pull/93/commits/55cc978c2a946944920292b2e6c4f51db939de81

ghatighorias:
Refactoring seems nice

PR #96, opened 2017-06-08
Adds holiday validation

ghatighorias:


web/models/terms/term.ex
```diff
@@ -33,8 +33,30 @@ defmodule CoursePlanner.Terms.Term do
...
+    st = changeset |> Changeset.get_field(:start_date) |> Date.cast!
+    en = changeset |> Changeset.get_field(:end_date) |> Date.cast!
+
+    validated_changesets =
+      holidays_changesets
+      |> Enum.map(fn(holiday_changeset) ->
+           hl = holiday_changeset.changes.date |> Date.cast!
+           case {Date.compare(st, hl), Date.compare(en, hl)} do
+             {:gt,   _} -> Changeset.add_error(holiday_changeset, :date,
+                                         "This holiday does not fall in the term range")
```
ggpasqualino: what do you think of having different messages for each case? Like "is before the term beginning" and "is after term end"?

PR #97, opened 2017-06-09
Removes status and status timestamps from term

ghatighorias:


web/controllers/term_controller.ex
```diff
@@ -76,6 +76,10 @@ defmodule CoursePlanner.TermController do
         conn
         |> put_status(404)
         |> render(CoursePlanner.ErrorView, "404.html")
+      {:error, _changeset} ->
+        conn
```
ggpasqualino: this should rather return a 500 status code and render the "500.html" template

ghatighorias:
This is amazing, I really don't get what is happening, now after merging with master the coverage is lower :(

rhnonose:
Makes sense if you remove code that is covered, but I wouldn't worry too much.

ghatighorias:
I can do that but it's in anycase a condition. if we could mock the model I could cover it but seems that need another library. @ggpasqualino  what is your opinion ?

ggpasqualino:
@ghatighorias I wouldn't worry too much with a -0.2% decrease, besides that, in this case if the `web/models/terms.ex` is well covered then we should be fine

test/models/term_test.exs
```diff
@@ -21,13 +20,41 @@ defmodule CoursePlanner.TermTest do
     refute changeset.valid?
   end
 
+  test "changeset with valid holidays" do
```
ggpasqualino: could you remove these two tests "changeset with valid holidays" and "changeset with invalid holidays"?

they shouldn't be here, this probably came from wrong merge

test/controllers/term_controller_test.exs
```diff
@@ -117,42 +136,55 @@ defmodule CoursePlanner.TermControllerTest do
     assert html_response(conn, 404)
   end
 
-  test "doesn't show deleted term", %{conn: conn} do
```
ggpasqualino: I don't understand why you removed this test

ghatighorias: hmm i mistook assuming we have a seperate test for "doesn't show non-existing term" and I removed this cause it was indirectly related to status. I rename this test and commit it back

ghatighorias: ok I checked. since the status is removed "doesn't show inexisting term" test already test the deleted scenario. here is the code
`test "doesn't show inexisting term", %{conn: conn} do
     conn = get conn, term_path(conn, :show, 1)
      assert html_response(conn, 404)
    end`

PR #98, opened 2017-06-09
Validate class date using term's date.

rhnonose:


web/models/class/class.ex
```diff
@@ -82,4 +85,27 @@ defmodule CoursePlanner.Class do
   end
   def validate_offered_course(changeset), do: changeset
 
+  defp validate_date(%{valid?: true, changes: %{date: date, offered_course_id: oc_id}} = changeset) do
```
ggpasqualino: is this working properly with class updates? it seems to me there wouldn't be a change on the field `offered_course_id` in an update

web/models/class/class.ex
```diff
@@ -82,4 +85,27 @@ defmodule CoursePlanner.Class do
...
+    |> Map.get(:start_date)
+    |> Date.cast!()
+
+    en = term
+    |> Map.get(:end_date)
+    |> Date.cast!()
+
+    case {Date.compare(st, date), Date.compare(en, date)} do
+      {:gt, _} -> Changeset.add_error(changeset, :date,
+                  "The class date is before term's beginning")
```
ggpasqualino: could you remove `The class date ` from the error message?

when this is added beside the field with the error tag in the page it doesn't read as nice

web/models/class/class.ex
rhnonose: You're probably right, I'll check.

web/models/class/class.ex
```diff
@@ -82,4 +83,27 @@ defmodule CoursePlanner.Class do
   end
   def validate_offered_course(changeset), do: changeset
 
+  defp validate_date(%{valid?: true, changes: %{date: date} = changes} = changeset) do
+    term = OfferedCourse
+    |> Repo.get(changes[:offered_course_id] || changeset.data.offered_course_id)
```
ggpasqualino: could you check if can use `Ecto.Changeset.get_field` instead of checking the internals of the changeset struct?

rhnonose: Yeah sorry, every time I forget about that function.

test/models/class_test.exs
```diff
@@ -1,14 +1,23 @@
...
 
-  @valid_attrs %{offered_course_id: nil, date: %{day: 17, month: 4, year: 2010}, finishes_at: %{hour: 14, min: 0, sec: 0}, starting_at: %{hour: 14, min: 0, sec: 0}}
+  @term_attrs %{name: "Term", start_date: "2010-01-01", end_date: "2010-12-31", status: "Active"}
+  @valid_course_attrs %{description: "some content", name: "some content", session_duration: %{hour: 14, min: 0, sec: 0}, status: "Planned"}
+  @valid_attrs %{offered_course_id: nil, date: %{day: 17, month: 4, year: 2010}, finishes_at: %{hour: 15, min: 0, sec: 0}, starting_at: %{hour: 14, min: 0, sec: 0}}
   @invalid_attrs %{}
+  @class_before_term %{offered_course_id: nil, date: %{day: 17, month: 4, year: 2009}, finishes_at: %{hour: 14, min: 0, sec: 0}, starting_at: %{hour: 14, min: 0, sec: 0}}
+  @class_after_term %{offered_course_id: nil, date: %{day: 17, month: 4, year: 2011}, finishes_at: %{hour: 14, min: 0, sec: 0}, starting_at: %{hour: 14, min: 0, sec: 0}}
+
+  defp create_course do
```
ggpasqualino: could you use the factory `insert(:offered_course)` instead of adding this function back?

PR #104, opened 2017-06-14
Check if class date is holiday, then fail

ghatighorias:


test/controllers/class_controller_test.exs
```diff
@@ -377,4 +377,24 @@ defmodule CoursePlanner.ClassControllerTest do
...
+                            holidays: [holiday]
+                           })
+
+    students = insert_list(3, :student)
+    teacher = insert(:teacher)
+    offered_course = insert(:offered_course, %{term: term1, course: course, students: students, teachers: [teacher]})
+
+    class_attrs = %{@valid_attrs | date: %{day: 1, month: 1, year: 2017}, offered_course_id: offered_course.id}
+    conn = post conn, class_path(conn, :create), class: class_attrs
+    assert class_path(conn, :new)
```
ggpasqualino: I don't understand this assertion here

test/models/class_test.exs
```diff
@@ -46,4 +46,12 @@ defmodule CoursePlanner.ClassTest do
     assert changeset.valid?
   end
 
+  test "class can't on holiday" do
```
ggpasqualino: class can't what on holiday? :D

web/models/class/class_helper.ex
```diff
@@ -5,9 +5,32 @@ defmodule CoursePlanner.ClassHelper do
...
+  alias Ecto.{Changeset, DateTime, Date}
   alias Ecto.Multi
 
+  def validate_for_holiday(%{valid?: true} = changeset) do
+    class_date = changeset |> Changeset.get_field(:date) |> Date.cast!
+    offered_course_id = changeset |> Changeset.get_field(:offered_course_id)
+
+    term = Repo.one(from t in Term,
+      join: oc in assoc(t, :offered_courses),
+      preload: [offered_courses: oc],
```
ggpasqualino: It doesn't seem like you need the offered courses later. Isn't the join enough?

web/models/class/class_helper.ex
```diff
@@ -5,9 +5,32 @@ defmodule CoursePlanner.ClassHelper do
...
+      where: oc.id == ^offered_course_id)
+
+    class_on_holiday? =
+      Enum.find(term.holidays, fn(holiday) ->
+        Date.cast!(holiday.date) == class_date
```
ggpasqualino: is this comparison reliable?

what about using `Date.compare(...) == :eq`?

web/models/class/class_helper.ex
```diff
@@ -5,9 +5,32 @@ defmodule CoursePlanner.ClassHelper do
...
+        end)
+
+    if class_on_holiday? do
+      add_error(changeset, :offered_course_id, "You cannot create a class on holiday")
```
ggpasqualino: I think it would make more sense to add the error to the `date` field of the class. What do you think?

web/models/class/class_helper.ex
ghatighorias: no cause without it complains that "oc" is an unbound variable

ggpasqualino: I am not sure I was clear, what I mean is the following
```elixir
term = Repo.one(from t in Term,
  join: oc in assoc(t, :offered_courses),
  where: oc.id == ^offered_course_id)
```
and I don't see why this would tell you that "oc" is an unbound variable

ghatighorias: i got your point  wrong

PR #105, opened 2017-06-14
Removes soft_deletion from Tasks

ghatighorias:


lib/course_planner/tasks.ex
```diff
@@ -2,17 +2,17 @@ defmodule CoursePlanner.Tasks do
...
+  alias Ecto.Changeset
   import Ecto.Query, except: [update: 2]
 
-  @tasks from t in Task, where: is_nil(t.deleted_at), preload: [:user]
-
   def all do
-    Repo.all(@tasks)
+    Task
+    |> Repo.all()
+    |> Repo.preload(:user)
```
rhnonose: I don't think you should preload a resource in a `all` call.
Make a separate query function to preload, put a parameter with preloading list or preload outside of the query.

ghatighorias: you mean like this ? 
`@tasks from t in Task, preload: [:user]

Repo.all(@tasks)`

ggpasqualino: @rhnonose this was there already, only in a different format

lib/course_planner/tasks.ex
```diff
@@ -2,17 +2,17 @@ defmodule CoursePlanner.Tasks do
...
   end
 
   def get(id) do
-    case Repo.one(from t in Task, where: is_nil(t.deleted_at) and t.id == ^id) do
+    case Repo.one(from t in Task, where: t.id == ^id) do
```
ggpasqualino: now that we only check the id we can use `Repo.get`

web/models/coherence/user.ex
```diff
@@ -28,7 +28,7 @@ defmodule CoursePlanner.User do
     field :active_at, :naive_datetime
     field :frozen_at, :naive_datetime
     field :graduated_at, :naive_datetime
-    has_many :tasks, Task
+    has_many :tasks, Task, on_delete: :nilify_all
```
ggpasqualino: 👍

lib/course_planner/tasks.ex
rhnonose: Damn, I'm the culprit then 😄

rhnonose: @ghatighorias I mean outside of `all/0` function when you call it or define a different one like `all_with_users/0` or `all(resource), do: Repo.all() |> Repo.preload(resource)`.

ghatighorias:
@ggpasqualino I'll increase the coverage a bit so plz wait for another commit :)

PR #106, opened 2017-06-14
Offered course permission

ggpasqualino:


lib/course_planner/permissions.ex
```diff
@@ -10,6 +12,13 @@ defimpl Canada.Can, for: CoursePlanner.User do
     when action in [:show, :index], do: true
   def can?(_role, _action, AttendanceController), do: false
 
+  def can?(%User{role: "Coordinator"}, _action, OfferedCourseController), do: true
```
rhnonose: This is not necessary as coordinators already have full permission per `def can?(%User{role: "Coordinator"}, _, _), do: true` and we won't remove it even if we lock the default permissions to false at the end.

ggpasqualino: you are right! good catch 🌴

PR #114, opened 2017-06-21
Laszlo.kardinal/offered course show page

laszlokardinal-dina:


priv/repo/seeds.exs
```diff
@@ -9,3 +9,43 @@ CoursePlanner.User.changeset(%CoursePlanner.User{},
     role: "Coordinator"},
     :seed)
 |> CoursePlanner.Repo.insert!
+
```
rhnonose: The `seed.exs` is supposed to run on staging/production as well. We discussed this issue a while before, and unless we split different seed files for dev/staging/prod, we generally shouldn't put test data into it.

PR #115, opened 2017-06-22
Ghatighorias/updates attendances when updating offered course

ghatighorias:
Moves the Attendance related functions to Attendance helpers
create the following functionality for updating the attendance records
 1 - check if any new student is added and create it's records
 2 - check if any student is removed and delete it's attendance records

This will be called from Offered_course update.


web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
...
+            not Enum.any?(offered_course.students, fn(student) ->
+                  attendance.student.id == student.id
+                end)
+          end)
+
+        %{
+          class_id: class.id,
+          missing_students: missing_students,
+          excessive_attendances: excessive_attendances
+         }
```
rhnonose: Can you fix the indentation? It's odd.

web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
...
+    query = from oc in OfferedCourse,
+    join: s in assoc(oc, :students),
+    join: c in assoc(oc, :classes),
+    join: cs in assoc(c, :students),
+    preload: [students: s, classes: {c, students: cs}]
+
+    offered_course_with_classes     = Repo.get(query, offered_course_id)
+    offered_course_with_attendances = get_course_attendances(offered_course_id)
+
+    cond do
```
rhnonose: This would be better with case and pattern matching, like:
```elixir
  case {att, cls} do
    {nil, nil} -> {:ok, nil}
    {nil, cls} -> ...
    {att, _}   -> ...
```

web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
...
+      is_nil(offered_course_with_attendances) && is_nil(offered_course_with_classes) ->
+        {:ok, nil}
+      is_nil(offered_course_with_attendances) ->
+        Enum.map(offered_course_with_attendances.classes, fn(class) ->
+          create_class_attendance_records(class)
```
rhnonose: You can call it like: `Enum.map(classes, &create_class_attendance_records/1)`

web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
...
+        %{
+          class_id: class.id,
+          missing_students: missing_students,
+          excessive_attendances: excessive_attendances
+         }
+    end)
+  end
+
+  defp fix_incorrect_attendance_record(incorrect_records)
+    when is_list(incorrect_records) do
```
rhnonose: Do you have a case where this is not a `list`? You should either remove this condition or add another call returning an error when it's not.

web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
 
     Repo.transaction(updated_multi)
   end
+
+  def update_class_attendance_records(offered_course_id) do
+    query = from oc in OfferedCourse,
+    join: s in assoc(oc, :students),
+    join: c in assoc(oc, :classes),
+    join: cs in assoc(c, :students),
```
ggpasqualino: Why do you join class with students? offered course students and class students are the same

web/models/attendance/attendance_helper.ex
```diff
@@ -84,4 +84,110 @@ defmodule CoursePlanner.AttendanceHelper do
...
+        |> fix_incorrect_attendance_record()
+    end
+  end
+
+  defp get_incorrect_attendance_record(offered_course) do
+    offered_course_classes = offered_course.classes
+
+    Enum.map(offered_course_classes, fn(class) ->
+      missing_students =
+        Enum.filter(offered_course.students, fn(student) ->
```
ggpasqualino: please use `Enum.reject` instead of filter and not

ggpasqualino:
This solution seems much simpler now 👍 

Could you add some tests?

web/models/attendance/attendance_helper.ex
ghatighorias: I have rewritten the whole solution. can you check the PR again ?

PR #120, opened 2017-07-03
Laszlo.kardinal/teacher show page

laszlokardinal-dina:


web/views/shared_view.ex
```diff
@@ -275,12 +283,19 @@ defmodule CoursePlanner.SharedView do
...
     render "class_list.html", classes: classes,
                               empty_text: empty_text,
                               show_attendances: show_attendances
   end
 
+  def course_list(offered_courses, opts \\ []) do
+    empty_text = opts[:empty_text] || "There are no courses here yet"
+
+    render "course_list.html", offered_courses: offered_courses,
+                               empty_text: empty_text
```
ggpasqualino: you are not using the `empty_text` on the template, is that intentional?

PR #123, opened 2017-07-03
Laszlo.kardinal/volunteer show page

laszlokardinal-dina:


web/views/shared_view.ex
```diff
@@ -275,12 +283,26 @@ defmodule CoursePlanner.SharedView do
...
+
+    render "course_list.html", offered_courses: offered_courses,
+                               empty_text: empty_text
+  end
+
+  def task_list(tasks, opts \\ []) do
+    empty_text = opts[:empty_text] || "There are no tasks here yet"
+
+    render "task_list.html", tasks: tasks,
+                             empty_text: empty_text
```
ggpasqualino: you are not using the `empty_text` in the template, is that on purpose?

PR #126, opened 2017-07-03
Laszlo.kardinal/term show page

laszlokardinal-dina:


web/views/shared_view.ex
```diff
@@ -275,12 +283,33 @@ defmodule CoursePlanner.SharedView do
...
+
+    render "task_list.html", tasks: tasks,
+                             empty_text: empty_text
+  end
+
+  def holiday_list(holidays, opts \\ []) do
+    empty_text = opts[:empty_text] || "There are no holidays here yet"
+
+    render "holiday_list.html", holidays: holidays,
+                                empty_text: empty_text
```
ggpasqualino: you are not using the `empty_text` on the template, is that on purpose?

laszlokardinal-dina: Yes. It has a default empty text. It can be specialized (e. g. this student has no course yet), but I can't tell if it is necessary. It's rather a UX question.

PR #128, opened 2017-07-04
Removed user's show page

laszlokardinal-dina:


test/controllers/user_controller_test.exs
```diff
@@ -113,32 +91,6 @@ defmodule CoursePlanner.UserControllerTest do
     assert html_response(conn, 403)
   end
 
-  test "does not update chosen user for non coordinator use", %{conn: _conn} do
```
ggpasqualino: I think this test could stay

web/controllers/user_controller.ex
```diff
@@ -32,10 +23,11 @@ defmodule CoursePlanner.UserController do
 
     case Repo.update(changeset) do
       {:ok, user} ->
-        Users.notify_user(user, current_user, :user_modified, user_url(conn, :show, user))
+        user_url = user_show_url(user)
+        Users.notify_user(user, current_user, :user_modified, user_url)
```
ggpasqualino: the notification needs the full url, but `user_show_url` generates only the path

test/controllers/user_controller_test.exs
laszlokardinal-dina: Sorry, that was a mistake. I put it back

web/controllers/user_controller.ex
laszlokardinal-dina: I added CoursePlanner.Endpoint.url there

PR #129, opened 2017-07-04
Added page title & backlink to layout

laszlokardinal-dina:
* Added page_title functions to views
* Added backlink to layout
* Added title to layout that uses page_title function if it presents

### DIFF
--- a/web/templates/layout/app.html.eex
+++ b/web/templates/layout/app.html.eex
@@ -17,7 +17,21 @@
     <div class="mdl-layout mdl-js-layout mdl-layout--fixed-header mdl-layout--fixed-drawer">
       <header class="mdl-layout__header">
         <div class="mdl-layout__header-row">
-          <span class="mdl-layout-title">Course Planner</span>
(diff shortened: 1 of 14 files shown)


web/templates/layout/app.html.eex
```diff
@@ -17,7 +17,21 @@
...
+          <%= if length(@conn.path_info) > 1 do %>
+            <a
+              href="#" onclick="location.href = document.referrer; return false;"
+              class="mdl-button mdl-js-button mdl-button--icon layout-backlink"
+            >
+              <i class="material-icons">&#xE5C4;</i>
+            </a>
+          <% end %>
+          <span class="mdl-layout-title">
+            <%= if :erlang.function_exported(@view_module, :page_title, 0) do %>
```
rhnonose: You can make this a little better (my suggestion, there are other ways) by doing something like this in `web.ex` inside `def view`'s `quote`:
```
def page_title, do: "Course Planner"
defoverridable [page_title: 0]
```
and now every view will have a default title which can be overridden by defining `page_title/0`.

PR #130, opened 2017-07-05
Introduces the json api for the calendar view

ghatighorias:


web/controllers/coherence/json_login.ex
```diff
@@ -0,0 +1,12 @@
+defmodule CoursePlanner.JsonLogin do
+  @moduledoc """
+  This module implements a login callback for json connection
+  """
+  use CoursePlanner.Web, :controller
```
ggpasqualino: I think it would be better to import the 2 functions we need instead of all the things from controller here

- Plug.Conn.put_status/2
- Phoenix.Controller.render/2

web/controllers/calendar_controller.ex
```diff
@@ -0,0 +1,24 @@
...
+       date = Map.get(changeset.changes, :date) || Date.utc_today()
+
+       week_range =  CalenderHelper.get_week_range(date)
+       offered_courses = CalenderHelper.get_user_classes(current_user, my_classes, week_range)
+       render conn, "index.json", offered_courses: offered_courses
+     false ->
+       conn
+       |> Map.put(:errors, changeset.errors)
+       |> put_status(406)
+       |> render(CoursePlanner.ErrorView, "406.json")
```
ggpasqualino: I would rather call `render(CoursePlanner.ErrorView, "406.json", errors: changeset.errors)` instead of assigning errors to the connection

web/controllers/calendar_controller.ex
```diff
@@ -0,0 +1,24 @@
...
+  use CoursePlanner.Web, :controller
+
+  alias CoursePlanner.CalenderHelper
+
+  def index(%{assigns: %{current_user: current_user}} = conn, params) do
+    changeset = CalenderHelper.validate(params)
+
+    case changeset.valid? do
+     true ->
+       my_classes = Map.get(changeset.changes, :my_classes) || false
```
ggpasqualino: Please check Ecto.Changeset.get_field/3

ghatighorias: done

web/controllers/calendar_controller.ex
ghatighorias: doesn't work cause the render function gets three argument. the assign parameter seems to be automatically populated by conn

here is the error : "no function clause matching in Plug.Conn.put_private/3"

web/controllers/coherence/json_login.ex
ghatighorias: done

web/controllers/calendar_controller.ex
ggpasqualino: https://hexdocs.pm/phoenix/Phoenix.Controller.html#render/4

ghatighorias: cool thanks.
I completely missing the keyword list. and that was the main issue

ghatighorias: done

test/controllers/calendar_controller_test.exs
```diff
@@ -0,0 +1,432 @@
...
+                          "finishes_at" => "12:00:00",
+                          "starting_at" => "10:00:00",
+                          "course_name" => "english",
+                          "teachers" => [%{"family_name" => "tf1",
+                                           "nickname" => "tn1",
+                                           "name" => "t1"},
+                                         %{"family_name" => "tf2",
+                                            "nickname" => "tn2",
+                                            "name" => "t2"}],
+                          "term_name" => "term1"}]}
```
rhnonose: Can you extract this expected result? It's probably occupying half of the file's lines and as I skimmed through, looks mostly identical in all tests.

web/models/calendar_helper.ex
```diff
@@ -0,0 +1,67 @@
+defmodule CoursePlanner.CalenderHelper do
```
rhnonose: Typo on this definition.

web/controllers/calendar_controller.ex
```diff
@@ -0,0 +1,25 @@
+defmodule CoursePlanner.CalendarController do
+  use CoursePlanner.Web, :controller
+
+  alias CoursePlanner.CalenderHelper
+  alias Ecto.Changeset
+
+  def index(%{assigns: %{current_user: current_user}} = conn, params) do
+    changeset = CalenderHelper.validate(params)
+
+    case changeset.valid? do
```
rhnonose: You can have a cleaner structure if you pattern match `valid?` deconstructing `changeset`:
```
case CalenderHelper.validate(params) do
  %{valid?: true} = changeset -> do stuff
  %{errors: errors} -> do stuff
end
```

ghatighorias: But in both case I need the changeset cause I use changeset.errors

test/controllers/calendar_controller_test.exs
ghatighorias: done

web/models/calendar_helper.ex
ghatighorias: doen

web/controllers/calendar_controller.ex
ghatighorias: done

test/controllers/calendar_controller_test.exs
```diff
@@ -0,0 +1,291 @@
...
+    test "fails when my_classes is not a boolean", %{conn: conn} do
+      params = %{date: "2017-01-01", my_classes: "this is not a boolean"}
+      conn = get conn, calendar_path(conn, :index), params
+      assert json_response(conn, 406) == %{"error" => [%{"my_classes" => "is invalid"}]}
+    end
+
+    @tag user_role: :coordinator
+    test "fails and shows all errors when both params are not formatted correctly", %{conn: conn} do
+      params = %{date: "wrong date format", my_classes: "this is not a boolean"}
+      errors = %{"error" => [%{"date" => "is invalid"}, %{"my_classes" => "is invalid"}]}
```
ggpasqualino: a list of maps is a bit weird here, what do you think of returning the following on parameter error?
```elixir
%{"errors" => %{"date" => "is invalid", "my_classes" => "is invalid"}}
```

ghatighorias: I put everything to Assign.errors so then i suggest one of the followings

1 - put "%{"errors" => %{"date" => "is invalid", "my_classes" => "is invalid"}}" in assign.errors but then what it shows will be %{"date" => "is invalid", "my_classes" => "is invalid"} 

2 - put them in another variable like assign.api_errors and then it can show 
%{"errors" => %{"date" => "is invalid", "my_classes" => "is invalid"}} when I return content of assign.api_error

@laszlokardinal-dina any opinion on this ?

ggpasqualino: @ghatighorias I don't understand why you are listing all this options

what I am saying here is that probably the format of the response on error would be better as following
```js
{
   "errors":{
      "date":"is invalid",
      "my_classes":"is invalid"
   }
}
```
instead of
```js
{
   "errors":[
      {
         "date":"is invalid"
      },
      {
         "my_classes":"is invalid"
      }
   ]
}
```

web/router.ex
```diff
@@ -20,6 +22,15 @@ defmodule CoursePlanner.Router do
     plug Coherence.Authentication.Session, protected: true
   end
 
+  pipeline :protected_api do
+    plug :accepts, ["json"]
+    plug :fetch_session
+    plug :fetch_flash
```
ggpasqualino: I think we don't need the plugs :fetch_flash, :protect_from_forgery, and :put_secure_browser_headers for this pipeline, could you check it please?

web/router.ex
```diff
@@ -31,6 +42,12 @@ defmodule CoursePlanner.Router do
   end
 
   scope "/", CoursePlanner do
+    pipe_through :protected_api
+
+    resources "/calendar", CalendarController, only: [:index]
```
ggpasqualino: conceptually I think it makes more sense to have
```elixir
resources "/calendar", CalendarController, only: [:show], singleton: true
```
because you don't get a list of calendars

test/controllers/calendar_controller_test.exs
ghatighorias: what I mean is that I already put the errors in Conn.assign.errors, then If I put errors in errors isn't it gonna be a bit weird ?
but anyhow I'll do it, just the naming was strange for me. please do ignore my previous comment

web/router.ex
ghatighorias: Done,

web/router.ex
ghatighorias: done. they were indeed excessive

test/controllers/calendar_controller_test.exs
ghatighorias: done

PR #131, opened 2017-07-05
Laszlo.kardinal/attendances

laszlokardinal-dina:


web/templates/attendance/fill_course_attendance.html.eex
```diff
@@ -1,34 +1,99 @@
...
+                    <tr class="attendance-date-table__thead-tr">
+                      <th class="attendance-date-table__th">Class Date</th>
+                    </tr>
+                  </thead>
+                  <tbody>
+                    <%= for class <- @offered_course.classes do %>
+                      <tr class="attendance-date-table__row">
+                        <td class="attendance-date-table__col">
+                          <%= class.date %>
+                          <%= String.slice Ecto.Time.to_iso8601(class.starting_at), 0, 5 %>
```
rhnonose: You can use a formatter like `Timex.format(class.starting_at, "{h24}:{m}")` for better formatting, also in case you want to change in the future.

web/views/attendance_view.ex
```diff
@@ -2,8 +2,9 @@ defmodule CoursePlanner.AttendanceView do
   use CoursePlanner.Web, :view
 
   def get_teacher_display_name(offered_course_teachers) do
-    Enum.reduce(offered_course_teachers, "", fn(teacher, out) ->
-              Enum.join([out, teacher.name, teacher.family_name], " ") end)
+    offered_course_teachers
+    |> Enum.map(fn(teacher) -> Enum.join([teacher.name, teacher.family_name], " ") end)
```
rhnonose: Could be cleaner like this: `|> Enum.map(&("#{&1.name} #{&1.family_name}"))`

ghatighorias: there is already a function name get_student_display_name. you can make it generic like "get_user_display_name" and use it for both

PS : in we have to modify the function to include the nick name if it's not "nil" too. cause in some places we now display the nickname and in some places we don't

web/templates/attendance/show_student.html.eex
```diff
@@ -1,23 +1,95 @@
...
+                          <% end %>
+                        </div>
+                      </th>
+                    </tr>
+                  </thead>
+                  <tbody>
+                    <%= for attendance <- @attendances do %>
+                      <tr class="attendance-table__row">
+                        <td class="attendance-table__col">
+                          <%= if attendance.attendance_type === "Present" do %>
```
ghatighorias: Previously we were thinking to extend the attendance types to be customized and with this strict if then we won't have such functionality

web/templates/attendance/fill_course_attendance.html.eex
ghatighorias: I would in any case extract this from form to the attendance_view

web/templates/attendance/show_student.html.eex
laszlokardinal-dina: If we plan to use arbitrary attendance types, we can't use icons, and we are stick with the dropdown solution, which is not the best solution in my opinion. I'm fine with extending the set of attendance types to a fixed number of types, this way we can associate them with icons. We need to discuss what other types can occur, and if they are practical.

web/views/attendance_view.ex
laszlokardinal-dina: I'll make a generic display name helper, and replace them in the code, because we really have a bunch of different solutions for that right now.

web/templates/attendance/fill_course_attendance.html.eex
laszlokardinal-dina: I already used this in the class index list. I think I tried Timex when I do that, but there where some type issues if I remember correctly, but I'm not sure. I will check it again. If I can get it work I will replace it at both places

web/templates/attendance/show_student.html.eex
ghatighorias: Yes I think we have to talk about it cause reducing the topic to absent and present is too simplistic

web/templates/attendance/fill_course_attendance.html.eex
laszlokardinal-dina: done

PR #139, opened 2017-07-10
Laszlo.kardinal/list menu fix

laszlokardinal-dina:


web/templates/attendance/index_student.html.eex
```diff
@@ -34,16 +34,26 @@
             >
               <i class="material-icons">more_vert</i>
             </button>
-            <ul class="mdl-menu mdl-menu--bottom-right mdl-js-menu"
-                for="tr_menu_<%= offered_course.id %>"
+            <ul
+              class="
+                mdl-menu mdl-js-menu
+                <%=
+                  if index > 10 && index > length(@offered_courses)-4 do
```
ghatighorias: I think you can safely use "and" instead of &&

web/templates/attendance/index_coordinator.html.eex
```diff
@@ -32,8 +32,18 @@
             >
               <i class="material-icons">more_vert</i>
             </button>
-            <ul class="mdl-menu mdl-menu--bottom-right mdl-js-menu"
-                for="tr_menu_<%= offered_course.id %>"
+            <ul
+              class="
+                mdl-menu mdl-js-menu
+                <%=
+                  if index > 10 && index > length(@offered_courses)-4 do
```
ghatighorias: I don't exactly get what this condition do. would you be interested to extract the condition to a function that describes it's purpose?

ggpasqualino: that's a good point, the whole thing could be extracted to a function `menu_position_class` or something like that

laszlokardinal-dina: the problem was that the lowest list items had menus that went down, but they did not always have enough space, and they messed up with the margins and scroll position. that won't happen if there are not too much items (e.g. won't cover the full screen), that's why the 10 condition is there. the another condition is there because only the last 3 list item would have menus that goes up, because the other menus would have enough space

laszlokardinal-dina: The pretty solution would be to create a separated reusable list component, that we would use instead of copying the same html to everywhere, but that's not the case now. I can extract the condition, but I don't think that would help so much.

web/templates/attendance/index_student.html.eex
laszlokardinal-dina: I'll change them

web/templates/attendance/index_coordinator.html.eex
ggpasqualino: @laszlokardinal-dina cool, let's leave the refactor for later then

web/templates/attendance/index_student.html.eex
laszlokardinal-dina: I changed the operator

PR #140, opened 2017-07-10
Introduces CSV as a type for system_variable

ghatighorias:


ghatighorias:
@rhnonose I don't know why your commit is dismissed without me having a chance to see it. I just committed my test :(

would you mind leave your comment again ?

ggpasqualino:
it is cool in general, but I am not very comfortable calling this a CSV, because CSV is a much broader specification, what do you think of calling it `list` type instead?

web/models/setting/settings.ex
```diff
@@ -25,6 +25,18 @@ defmodule CoursePlanner.Settings do
     Repo.all(@all_query)
   end
 
+  def get_value(name) do
```
ggpasqualino: it is a nice interface to receive a default value which will be returned when `system_variable` matches to nil on https://github.com/digitalnatives/course_planner/pull/140/files#diff-46197ed733140459d99a5c8b132185ffR32 just like `Map.get` interface

```elixir
def get_value(name, default \\ nil) do
...
      nil -> {:ok, default}
...
```

ghatighorias:
list is fine with me. will change it in the next commit

web/models/setting/system_variable.ex
```diff
@@ -79,6 +73,26 @@ defmodule CoursePlanner.SystemVariable do
...
+      "list"     -> parse_list(value)
+      "integer" -> parse_integer(value)
+      "boolean" -> parse_boolean(value)
+      _         -> {:error, "unknown type"}
+    end
+  end
+
+  def parse_list(value) do
+    parse_list =
+    value
```
ggpasqualino: I am not sure it is more understandable, but you may replace these 4 lines by
```elixir
String.split(value, ~r/,\s*/, trim: true)
```
what do you think?

web/models/setting/settings.ex

web/models/setting/system_variable.ex
ghatighorias: If you say so :P
I was thinking of it then I though not everyone like regex, but honestly I have no preference

ggpasqualino: if it is a simple one it is fine, if it needs to get more complicate then leave it as it is :)

ghatighorias: Hmmmm after some test the problem is that for removing the leading white space and trailing one there should be another regex. Are we still interested in it ?

ghatighorias: However this will do the job "String.split(value, [" ", ","] , trim: true)"

PR #141, opened 2017-07-10
Do not show tasks that are expired/done.

rhnonose:


lib/course_planner/tasks.ex
```diff
@@ -47,22 +47,32 @@ defmodule CoursePlanner.Tasks do
     end
   end
 
-  def get_user_tasks(id) do
-    Repo.all(from t in Task, where: t.user_id == ^id, preload: [:user])
+  def get_for_user(id, now) do
+    Repo.all(from t in Task, where: t.user_id == ^id and t.finish_time > ^now, preload: [:user])
```
ggpasqualino: I understand hiding the past unassigned tasks, but why would we hide the assigned ones? what if I want to check which tasks I did last week for example?

web/controllers/task_controller.ex
```diff
@@ -12,8 +12,8 @@ defmodule CoursePlanner.TaskController do
 
   def index(%{assigns: %{current_user: %{id: id, role: "Volunteer"}}} = conn, _params) do
     render(conn, "index_volunteer.html",
-      available_tasks: Tasks.get_unassigned_tasks(),
-      your_tasks: Tasks.get_user_tasks(id))
+      available_tasks: Tasks.get_unassigned(Timex.now()),
```
ggpasqualino: what do you think about extracting the call to `Timex.now()` into a variable? the way it is now there might be a slightly difference between each call

rhnonose: You're right, I'll do that.

lib/course_planner/tasks.ex
rhnonose: Maybe have a filter for past tasks?
I believe it was to make the list clean but am not sure, @ghatighorias specified that.

ggpasqualino: I think it makes sense to split in the template so we would have 3 tables: available, future and past tasks

lib/course_planner/tasks.ex
```diff
@@ -47,22 +47,32 @@ defmodule CoursePlanner.Tasks do
...
+    Repo.all(from t in Task, where: is_nil(t.user_id) and t.finish_time > ^now)
   end
 
-  def grab(task_id, user_id) do
+  def grab(task_id, user_id, now) do
     case get(task_id) do
       {:ok, task} ->
         task
         |> Task.changeset()
+        |> validate_finish_time(now)
```
ggpasqualino: it would be good to validate also if the task was already assigned to someone else, let's say a volunteer leaves the task index page open the whole day then they decide to click the link to grab a task, this task may be assigned already in the meantime

lib/course_planner/tasks.ex
```diff
@@ -47,22 +47,41 @@ defmodule CoursePlanner.Tasks do
...
   end
 
-  def grab(task_id, user_id) do
-    case get(task_id) do
-      {:ok, task} ->
-        task
-        |> Task.changeset()
-        |> Changeset.put_change(:user_id, user_id)
-        |> Repo.update()
+  def get_past_tasks(id, now) do
```
ggpasqualino: can you add a test for this?

PR #143, opened 2017-07-10
Laszlo.kardinal/readd user show page

laszlokardinal-dina:


web/controllers/user_controller.ex
```diff
@@ -11,6 +11,22 @@ defmodule CoursePlanner.UserController do
     render(conn, "index.html", users: Users.all())
   end
 
+  def show(%{assigns: %{current_user: %{role: role}}} = conn, %{"id" => id}) do
+    if role === "Coordinator" do
```
rhnonose: This is not necessary as we are already restricting access to this controller in `permissions.ex`.
If you tested with other roles and you could access it, that's a bug and we should fix there, not here (it shouldn't even call this function).

web/templates/user/show.html.eex
```diff
@@ -1,31 +1,64 @@
...
+      col-xs-12
+      col-sm-offset-1 col-sm-10
+      col-md-offset-2 col-md-8
+      col-lg-offset-2 col-lg-8
+    ">
+      <div class="row middle-xs page-header">
+        <div class="col-xs-6 col-sm-9 col-md-10 page-title">
+          <%= CoursePlanner.SharedView.display_user_name(@user) %>
+        </div>
+        <%= if @conn.assigns.current_user.role == "user" do %>
```
rhnonose: Did you mean `Coordinator`? Or something else, but this looks wrong.

laszlokardinal-dina: I changed that, that was an accident, I used the coordinator show page, and I replaced that one too. I will update the PR soon

web/controllers/user_controller.ex
laszlokardinal-dina: I will use the permission.ex if I can find out how can I add rules to that. Yesterday I spent relatively long time to replace that condition, but I still couldn't do that. I can't tell if it's necessary to hide the user show page anyway.

laszlokardinal-dina: Every user had access to their own show page before, that's what I tried to stop. We won't redirect them to the user page ever

rhnonose: If you need help with `permissions.ex`, drop me a word, it's pattern matching galore so it might confusing at first sight.

web/templates/user/show.html.eex
laszlokardinal-dina: I updated the PR

web/controllers/user_controller.ex
laszlokardinal-dina: I could deal with the pattern matching, but I couldn't solve the issue, so now every user can see their own show page. Gui created a card for cleaning up permissions

test/controllers/user_controller_test.exs
```diff
@@ -108,6 +131,15 @@ defmodule CoursePlanner.UserControllerTest do
     assert html_response(conn, 403)
   end
 
+  test "shows the user himself" do
+    user = insert(:user)
+    user_conn = Phoenix.ConnTest.build_conn()
+    |> assign(:current_user, user)
```
ghatighorias: you don't need to duplicate this part. just use "login_as" and it will give you a "conn" object
then the user already is assigned to it. you can access it via "conn.assigns.current_user"

test/controllers/user_controller_test.exs
```diff
@@ -117,9 +149,19 @@ defmodule CoursePlanner.UserControllerTest do
     assert html_response(conn, 200) =~ "Foo Bar"
   end
 
+  test "coordinator updates his profile" do
+    user = insert(:coordinator)
+    user_conn = login_as(:coordinator)
+    |> assign(:current_user, user)
```
ghatighorias: same as the comment on line 137

web/controllers/user_controller.ex
```diff
@@ -11,6 +11,16 @@ defmodule CoursePlanner.UserController do
     render(conn, "index.html", users: Users.all())
   end
 
+  def show(%{assigns: %{current_user: %{role: role}}} = conn, %{"id" => id}) do
```
ghatighorias: the variable "role" in %{current_user: %{role: role}}} in unused

web/templates/user/show.html.eex
```diff
@@ -1,31 +1,64 @@
-<h2>Show user</h2>
```
ghatighorias: In case the coordinator is checking a student record, She should be allowed to see the student_id and other records you hide from student or other users

web/templates/user/show.html.eex
```diff
@@ -1,31 +1,64 @@
...
+      col-xs-12
+      col-sm-offset-1 col-sm-10
+      col-md-offset-2 col-md-8
+      col-lg-offset-2 col-lg-8
+    ">
+      <div class="row middle-xs page-header">
+        <div class="col-xs-6 col-sm-9 col-md-10 page-title">
+          <%= CoursePlanner.SharedView.display_user_name(@user) %>
+        </div>
+        <%= if @conn.assigns.current_user.role == "Coordinator" do %>
```
ghatighorias: Doesn't this line implies that the student cannot edit his own page ?

laszlokardinal-dina: The user can edit his own profile from the navigation bar

web/templates/user/show.html.eex
laszlokardinal-dina: Those fields belong to the student role and they are shown there already

web/controllers/user_controller.ex
laszlokardinal-dina: I'll fix it

laszlokardinal-dina: fixed

test/controllers/user_controller_test.exs
laszlokardinal-dina: fixed

test/controllers/user_controller_test.exs
laszlokardinal-dina: fixed

web/controllers/user_controller.ex
```diff
@@ -28,9 +38,15 @@ defmodule CoursePlanner.UserController do
       {:ok, user} ->
         Users.notify_user(user, current_user, :user_modified, user_show_url(user))
 
-        conn
-        |> put_flash(:info, "User updated successfully.")
-        |> redirect(to: dashboard_path(conn, :show))
+        if conn.assigns.current_user.role === "Coordinator" do
```
ggpasqualino: what do you think about checking `conn.assigns.current_user.id == id` instead?
my idea is that coordinators can see the message `"Your profile has been updated successfully."` if they edit themselves

PR #147, opened 2017-07-11
Adds comment for attendance and enforces them to be from settings

ghatighorias:
"ATTENDANCE_DESCRIPTORS" is used for the validation of the comments


test/models/attendance_test.exs
```diff
@@ -5,20 +5,48 @@ defmodule CoursePlanner.AttendanceTest do
...
+    {:ok, %{student: student, class: class}}
+  end
+
+  test "changeset with valid attributes", context do
+    changeset = Attendance.changeset(%Attendance{}, %{@valid_attrs | student_id: context.student.id, class_id: context.class.id})
     assert changeset.valid?
   end
 
-  test "changeset with invalid attributes" do
+  test "changeset with invalid attributes", _context do
```
rhnonose: You can omit this `_context`.

test/models/attendance_test.exs
```diff
@@ -5,20 +5,48 @@ defmodule CoursePlanner.AttendanceTest do
...
     changeset = Attendance.changeset(%Attendance{}, @invalid_attrs)
     refute changeset.valid?
   end
+
+  describe "attendance comment" do
+    test "passes when comment is empty", context do
```
rhnonose: It's better to pattern match the context like `%{student: student, class: class}` because if you change the `setup` the error is clearer.

ghatighorias: done

test/models/attendance_test.exs
ghatighorias: done

ghatighorias:
@rhnonose I think that was the idea. That next to the selected attendance there will be and option that can further elaborate the attendance based on the options available in the ATTENDANCE_DESCRIPTOR "list" options. And teacher or coordinator can select from them. The option shall be available for both absent and present.

@laszlokardinal-dina correct me if I am wrong.

web/models/attendance/attendance.ex
```diff
@@ -14,9 +17,32 @@ defmodule CoursePlanner.Attendance do
 
   def changeset(struct, params \\ %{}) do
     struct
-    |> cast(params, [:attendance_type, :class_id, :student_id])
+    |> cast(params, [:attendance_type, :class_id, :student_id, :comment])
     |> validate_required([:attendance_type, :class_id, :student_id])
     |> cast_assoc(:class)
     |> cast_assoc(:student)
+    |> validate_comment()
```
ggpasqualino: is it possible to use `Ecto.Cahngeset.validate_inclusion` instead of creating a new function?

priv/repo/seeds.exs
```diff
@@ -57,3 +57,14 @@ CoursePlanner.Repo.delete_all CoursePlanner.User
       visible: true
     })
 |> CoursePlanner.Repo.insert!
+
+%CoursePlanner.SystemVariable{}
+|> CoursePlanner.SystemVariable.changeset(
+    %{
+      key: "ATTENDANCE_DESCRIPTORS",
```
ggpasqualino: what do you think of "ATTENDANCE_DESCRIPTIONS" instead? `DESCRIPTORS` sounds kinda technical to me

ghatighorias: I though descriptor suits better, but I am fine with both. but double check the meaning and see if you still "descriptions" more relevant.

web/models/attendance/attendance.ex
ghatighorias: yes, it will look like this 
`
struct
    |> cast(params, [:attendance_type, :class_id, :student_id, :comment])
    |> validate_required([:attendance_type, :class_id, :student_id])
    |> cast_assoc(:class)
    |> cast_assoc(:student)
    |> validate_inclusion(:comment, Settings.get_value("ATTENDANCE_DESCRIPTORS"))`

I found it ugly and that's why i wrote another function, but if it's good I change it to this one

priv/repo/seeds.exs
ghatighorias: done

web/models/attendance/attendance.ex
ghatighorias: done

priv/repo/seeds.exs
ggpasqualino: > but double check the meaning and see if you still "descriptions" more relevant.

I did it before I suggested it

web/models/attendance/attendance.ex
ggpasqualino: you may extract `Settings.get_value("ATTENDANCE_DESCRIPTIONS")` to a function in order to make it nicer, but we shouldn't reinvent Ecto validations

PR #148, opened 2017-07-11
Add task sorting by updated_at and finish_time.

rhnonose:


lib/course_planner/tasks.ex
```diff
@@ -47,14 +47,26 @@ defmodule CoursePlanner.Tasks do
     end
   end
 
-  def get_user_tasks(id) do
-    Repo.all(from t in Task, where: t.user_id == ^id, preload: [:user])
+  def get_user_tasks(id, sort_opt) do
+    sort_opt
+    |> task_query()
+    |> where([t], t.user_id == ^id)
```
ghatighorias: This is super nice. I never knew one can pipe query. I like it

lib/course_planner/tasks.ex
```diff
@@ -47,14 +47,26 @@ defmodule CoursePlanner.Tasks do
...
-  def get_unassigned_tasks do
-    Repo.all(from t in Task, where: is_nil(t.user_id))
+  def get_unassigned_tasks(sort_opt) do
+    sort_opt
+    |> task_query()
+    |> where([t], is_nil(t.user_id))
+    |> Repo.all()
   end
 
+  defp task_query(sort_opt), do: Task |> sort(sort_opt) |> preload(:user)
```
ggpasqualino: could you create tests for this ordering?

ggpasqualino: could you create tests for this ordering?

PR #149, opened 2017-07-13
Laszlo.kardinal/calendar ui

laszlokardinal-dina:


web/static/js/calendar.js
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  function renderHour ( hour ) {
+    return `0${hour}:00`.slice( -5 );
+  }
+
+  function renderCalendar ( startDate, classes, calendar ) {
+    const monday = new Date( startDate );
+
+    let days = new Array( 5 ).fill( monday ).map(
```
ggpasqualino: Does that mean we are displaying from Monday to Friday? I think we should include the whole week, the weekend is the most important for OLIve if I am not wrong.

@ghatighorias is my assumption correct?

PR #156, opened 2017-07-19
Fixes bug that caused edit always shows the default time and date

ghatighorias:
The bug is because I put the default value to priority and do not check for the changed value of changeset.

### DIFF
--- a/web/views/shared_view.ex
+++ b/web/views/shared_view.ex
@@ -110,7 +110,7 @@ defmodule CoursePlanner.SharedView do
   end
 
   def form_date(form, field, opts \\ []) do
-    value = Map.get(form.source.changes, field) || opts[:value] || Ecto.Date.utc()
+    value = input_value(form, field) || opts[:value] || Ecto.Date.utc()
     class = opts[:class] || ""
     label = opts[:label] || humanize(field)
     {error, _} = Keyword.get form.errors, field, {nil, nil}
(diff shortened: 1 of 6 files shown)


ghatighorias:
The changed value of a changeset is an Ecto type and that does not have field name "minute" neither the function "utc_now" so I had to replace them with "min" and "utc" and use Ecto types for date, time and datetime

web/views/shared_view.ex
```diff
@@ -110,7 +110,7 @@ defmodule CoursePlanner.SharedView do
   end
 
   def form_date(form, field, opts \\ []) do
-    value = opts[:value] || Date.utc_today()
+    value = Map.get(form.source.changes, field) || opts[:value] || Ecto.Date.utc()
```
ggpasqualino: what do you think of using
```elixir
Phoenix.HTML.Form.input_value(form, field)
```
instead of
```elixir
Map.get(form.source.changes, field)
```

web/views/shared_view.ex
```diff
@@ -144,10 +144,10 @@ defmodule CoursePlanner.SharedView do
   end
 
   def form_datetime(form, field, opts \\ []) do
-    value =
+    value = Map.get(form.source.changes, field) ||
```
rhnonose: Can you extract this in a function? This `case` after a `||` looks very confusing.

web/views/shared_view.ex
ggpasqualino: inside the SharedView we should be able to call
```elixir
input_value(form, field)
```
instead of
```elixir
Phoenix.HTML.Form.input_value(form, field)
```
though

web/views/shared_view.ex
ghatighorias: I simplified it but basically when you wanna merge two map it won't accept nil as input, that's why i wrote the case,

web/views/shared_view.ex
ghatighorias: done

PR #157, opened 2017-07-20
Add validation for user comment length

ggpasqualino:


ghatighorias:
hmm, is it so necessary to restrict the comment on a student to 255 ? cause we can actually extend it to a :text field and let the coordinator to update it as they wish

Nevertheless, The code itself looks perfect

ggpasqualino:
@ghatighorias it is already limited to 255 by the database, I added an application validation in order to have a proper error when user gives a comment bigger than that instead of a 500 error

PR #161, opened 2017-07-25
Add max volunteer per task

ghatighorias:


lib/course_planner/permissions.ex
```diff
@@ -7,9 +7,12 @@ defimpl Canada.Can, for: CoursePlanner.User do
     PageController,
     ScheduleController,
     TaskController,
-    User,
```
rhnonose: I prefer leaving lists that are frequently modified with the dangling `,` for better diffs 😄

web/controllers/task_controller.ex
```diff
@@ -57,30 +63,43 @@ defmodule CoursePlanner.TaskController do
...
+    with {:ok, task} <- Tasks.get(id)
+     do
+       changeset =
+         task
+         |> Repo.preload([:volunteers])
+         |> Task.changeset(task_params)
+
+       volunteer_ids = Map.get(task_params, "volunteer_ids", [])
+       volunteers = Repo.all(from v in Volunteers.query(), where: v.id in ^volunteer_ids)
+       changeset = Task.update_volunteer(changeset, :volunteers, volunteers)
```
rhnonose: This block of code looks like the same as in `TaskController.create/2`, maybe you can extract them?

web/controllers/task_controller.ex
```diff
@@ -57,30 +63,43 @@ defmodule CoursePlanner.TaskController do
...
-        |> redirect(to: task_path(conn, :show, task))
-      {:error, :not_found} ->
-        conn
-        |> put_status(404)
-        |> render(CoursePlanner.ErrorView, "404.html")
-      {:error, task, changeset} ->
-        render(conn, "edit.html", task: task, changeset: changeset, users: Volunteers.all())
-    end
+  def update(conn, %{"id" => id, "task" => task_params})do
+    with {:ok, task} <- Tasks.get(id)
```
rhnonose: I don't think you need a `with` clause when there's only one statement, `case` should suit better.

web/controllers/task_controller.ex
```diff
@@ -100,23 +119,40 @@ defmodule CoursePlanner.TaskController do
...
 
-  @error_messages %{
-    not_found: "Task was not found.",
-    already_finished: "Cannott grab as task is already finished.",
-    already_assigned: "Cannot grab as task is already assigned."
-  }
-
-  def grab(%{assigns: %{current_user: %{id: user_id}}} = conn, %{"task_id" => task_id}) do
-    case Tasks.grab(task_id, user_id, Timex.now()) do
+  def grab(%{assigns: %{current_user: %{id: volunteer_id}}} = conn, %{"task_id" => task_id}) do
```
rhnonose: `grab/2` and `drop/2` looks exactly the same apart from `Tasks.grab/2` and `Tasks.drop/2` calls, maybe you can extract it.

web/models/tasks/task.ex
```diff
@@ -21,6 +27,47 @@ defmodule CoursePlanner.Tasks.Task do
...
+    else
+      changeset
+    end
+  end
+  defp validate_volunteers_limit(changeset), do: changeset
+
+  defp validate_datetime(%{valid?: true} = changeset) do
+    finish_time = Changeset.get_field(changeset, :finish_time)
+    start_time = Changeset.get_field(changeset, :start_time)
+    now = Timex.now()
```
rhnonose: I don't like this `Timex.now()` call here. Maybe the caller should call it and pass as a parameter.

ghatighorias: Why is it a bad thing ?

web/controllers/task_controller.ex
ghatighorias: hmm i understand they are very close to each other but i wanted to be as explicit as possible.
@ggpasqualino what is your idea on this ?

web/controllers/task_controller.ex
ghatighorias: there is already a case controlling the update result. another one would make it more messy

rhnonose: Maybe have the inside `case` in a different function? The function body is already pretty big with nested statements, so it should be simplified in some way.

web/models/tasks/task.ex
rhnonose: Makes the function inpure, meaning if you take a past task and make a copy, for example, it's going to break the validation. I just don't like having side-effects inside a schema validation function overall.
It'll break the validation when updating a task when it's past the finish time, also (no sure if it's an intended behavior to make the task "immutable" after the finish time).

web/controllers/task_controller.ex
ghatighorias: done

lib/course_planner/permissions.ex
ghatighorias: no objection.
it's just on my nerve ( speaking of being a c++ programmer . hhahahha)

ghatighorias: done

lib/course_planner/tasks.ex
```diff
@@ -47,59 +23,75 @@ defmodule CoursePlanner.Tasks do
...
-      1  -> {:ok, changeset}
-      -1 -> {:error, :already_finished}
-      _  -> {:error, :unkown}
+  def drop(task_id, volunteer_id) do
+    with {:ok, task} <- get(task_id),
+        %{valid?: true} = changeset <- Task.changeset(task)
+     do
+      drop_volunteer = Volunteers.get!(volunteer_id)
+      updated_volunteer_list = List.delete(task.volunteers, drop_volunteer)
+
```
ghatighorias: hmmm
the second statement never crashes, only if the first statement is not a list
the first statement will crash only in case volunteer is removed ( but is it necessary to handle a case like this : e.g. coordinator removes a volunteer exactly at the a time that he is grabbing a task

rhnonose: That's kinda the point with `with`statements. You can match different errors as well.

lib/course_planner/tasks.ex
```diff
@@ -47,59 +23,81 @@ defmodule CoursePlanner.Tasks do
...
   end
 
   def get_past(sort_opt, id, now) do
     sort_opt
     |> task_query()
-    |> where([t], t.user_id == ^id)
     |> where([t], t.finish_time < ^now)
     |> Repo.all()
+    |> Enum.reject(fn(task) ->
+         not Enum.any?(task.volunteers, &(&1.id == id))
```
ggpasqualino: please avoid negative conditions, positive conditions are usually more understandable
```elixir
...
|> Enum.filter(fn(task) ->
        Enum.any?(task.volunteers, &(&1.id == id))
    end)
```

priv/repo/migrations/20170719103736_add_max_volunteer_per_task.exs
```diff
@@ -0,0 +1,10 @@
+defmodule CoursePlanner.Repo.Migrations.AddMaxVolunteerPerTask do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tasks) do
+      remove :user_id
+      add :max_volunteers, :integer
```
ggpasqualino: you added this as a required field in the ecto schema, we should then add `not null` to this migration as well, and in order to not break the migration in the staging server we also should have a default value (1 is fine to match the previous behavior)

web/templates/task/index.html.eex
```diff
@@ -33,8 +34,13 @@
             <%= String.slice NaiveDateTime.to_iso8601(task.finish_time), 11, 5 %>
           </td>
           <td class="mdl-data-table__cell--non-numeric">
-            <%= if task.user_id do %>
-              <%= CoursePlanner.SharedView.display_user_name(task.user) %>
+            <%= task.max_volunteers %>
+          </td>
+          <td class="mdl-data-table__cell--non-numeric">
+            <%= if task.volunteers do %>
```
ggpasqualino: is this ever falsy? as far as I remember this will be always a list, an empty list in case there is no volunteers?

web/views/task_view.ex
```diff
@@ -1,14 +1,16 @@
...
-    else
-      "no one"
-    end
+  def task_user(conn, volunteers) when not is_nil(volunteers) do
+    volunteers
+    |> Enum.map(fn(volunteer) ->
+         link volunteer.name, to: volunteer_path(conn, :show, volunteer)
+       end)
   end
+  def task_user(_conn, _volunteers), do: "no one"
```
ggpasqualino: is this ever reached? as far as I remember volunteers will be always a list, an empty list in case there is no volunteers

web/views/task_view.ex
```diff
@@ -1,14 +1,16 @@
...
+  alias CoursePlanner.{SharedView, Volunteers}
+  alias Ecto.Changeset
 
-  def task_user(conn, task) do
-    if task.user_id do
-      link task.user.name, to: volunteer_path(conn, :show, task.user)
-    else
-      "no one"
-    end
+  def task_user(conn, volunteers) when not is_nil(volunteers) do
```
ggpasqualino: is this shown properly? maybe it is worth to join the links with a comma or newline

lib/course_planner/tasks.ex
```diff
@@ -47,59 +23,81 @@ defmodule CoursePlanner.Tasks do
...
+      |> Task.drop_volunteer(:volunteers, updated_volunteer_list)
+      |> Repo.update()
+    else
+      error -> error
     end
   end
 
-  defp validate_already_assigned(changeset, %{user_id: nil}), do: {:ok, changeset}
-  defp validate_already_assigned(_changeset, _), do: {:error, :already_assigned}
+  def update_changeset_volunteers(changeset, task_params, volunteer_param_name) do
```
ggpasqualino: do we need this `volunteer_param_name`? it seems it is always the same

web/models/tasks/task.ex
```diff
@@ -21,6 +27,47 @@ defmodule CoursePlanner.Tasks.Task do
     struct
     |> cast(params, @cast_params)
     |> validate_required(@required_params)
-    |> assoc_constraint(:user)
+    |> validate_datetime()
+    |> validate_number(:max_volunteers, greater_than: 0, less_than: 1_000)
   end
+
+  def drop_volunteer(changeset, field, field_data), do: put_assoc(changeset, field, field_data)
+  def update_volunteer(changeset, field, field_data) do
```
ggpasqualino: do we need the `field` parameter? it seems it is always `volunteers` and we already have this information on the function name

web/controllers/task_controller.ex
ggpasqualino: Previously the updating logic was encapsulated into the `Tasks.update/2` function, I think that was an appropriate solution, so we could just adapt it to accept multiple volunteers instead of one

priv/repo/migrations/20170719103736_add_max_volunteer_per_task.exs
ghatighorias: good point, didn't cross my mind

web/models/tasks/task.ex
ggpasqualino: I don't get why this validation with the current date is necessary

lib/course_planner/tasks.ex
```diff
@@ -47,59 +23,81 @@ defmodule CoursePlanner.Tasks do
...
-  def grab(task_id, user_id, now) do
-    with {:ok, task}              <- get(task_id),
-      %{valid?: true} = changeset <- Task.changeset(task),
-      {:ok, changeset}            <- validate_finish_time(changeset, now),
-      {:ok, changeset}            <- validate_already_assigned(changeset, task)
-    do
+  def grab(task_id, volunteer_id) do
+    with {:ok, task} <- get(task_id),
+        %{valid?: true} = changeset <- Task.changeset(task)
+     do
```
ghatighorias: cause it will be done in the changeset as part of the update in anycase

web/models/tasks/task.ex
ghatighorias: that's true, i just wanted to write it generic :P

lib/course_planner/tasks.ex
ghatighorias: yes, i just didn't want tasks to be dependent on the parameters, i thought it's already controller is dependent on it, and tasks does not need to be aware of that. shall I make it explicit ?

web/models/tasks/task.ex
ghatighorias: done

lib/course_planner/tasks.ex
ggpasqualino: then it is probably more interesting to receive volunteer_ids straight from the controller instead of params and the param name, because this way you still depend on the params here

lib/course_planner/tasks.ex
ggpasqualino: okay, in this case there should be a validation if this user is already assigned, so we don't have the same volunteer assigned multiple times to a task

ghatighorias: but we won't cause the changeset will not consider it as an update.

ghatighorias: alternatively I can use an Enum.uniq to remove the duplicates as well. any good idea ?

lib/course_planner/tasks.ex
ghatighorias: i had it there, but i extract it to this function due to similarity.
I can create a function in controller and have it there, is this a better idea ?

lib/course_planner/tasks.ex
ghatighorias: done

lib/course_planner/tasks.ex
ghatighorias: done

priv/repo/migrations/20170719103736_add_max_volunteer_per_task.exs
ghatighorias: done

web/models/tasks/task.ex
ghatighorias: This validation assure that "finish_time" is always after the current time. so you cannot create a task with start and finish in the past. it enforces the task to be finished in the future

ghatighorias: it additionally stop you to grab or drop tasks that are already finished, unless coordinator extend its finish_time

ggpasqualino: that's fine for create, grab, and drop 👍  
but as @rhnonose already mentioned this will also prevent any updates to finished tasks unless the coordinator changes the finish time as well

web/controllers/task_controller.ex
ggpasqualino: for now I don't see any problem extracting the response formatting to another function

lib/course_planner/tasks.ex
```diff
@@ -47,59 +23,80 @@ defmodule CoursePlanner.Tasks do
...
   end
 
   def get_for_user(sort_opt, id, now) do
     sort_opt
     |> task_query()
-    |> where([t], t.user_id == ^id)
     |> where([t], t.finish_time > ^now)
     |> Repo.all()
+    |> Enum.reject(fn(task) ->
+         not Enum.any?(task.volunteers, &(&1.id == id))
```
ggpasqualino: change this negative condition also please

lib/course_planner/tasks.ex
ggpasqualino: > but the changeset will not consider it as an update if you have the same member twice. it is not a change

I don't see any reason for this to happen

> alternatively I can use an Enum.uniq to remove the duplicates as well. any good idea ?

I think it is okay

lib/course_planner/tasks.ex
ggpasqualino: I got lost...

what I mean is to change this function signature to 
```elixir
def update_changeset_volunteers(changeset, volunteer_ids) do
```
then call it like
```elixir
Tasks.update_changeset_volunteers(changeset, task_params["volunteer_ids"])
```

web/models/tasks/task.ex
ghatighorias: then i can create another changeset just for coordinator update

ghatighorias: done

lib/course_planner/tasks.ex
ghatighorias: done

web/views/task_view.ex
ghatighorias: I completely removed this function cause it was creating a link to other volunteer show page which is forbidden to be accessed by other volunteers

web/views/task_view.ex
ghatighorias: fixed

web/templates/task/index.html.eex
ghatighorias: fixed

lib/course_planner/tasks.ex
ghatighorias: I added the Enum.uniq and a test for this. but in any case even without the enum.uniq it was not adding a duplicated relation.

lib/course_planner/tasks.ex
ghatighorias: calling it this way, will cause a crash when the parameter does not exist.
I moved "volunteer_ids = Map.get(task_params, "volunteer_ids", [])" this to the controller and change "update_changeset_volunteers" function so it can be called like :

"Tasks.update_changeset_volunteers(changeset, volunteer_ids)

web/controllers/task_controller.ex
ghatighorias: okay, I extracted it to a private function to handle both grab and drop together

web/controllers/task_controller.ex
ghatighorias: I created a update again. see if it is good

rhnonose: SGTM

web/controllers/task_controller.ex
```diff
@@ -100,23 +103,36 @@ defmodule CoursePlanner.TaskController do
     end
   end
 
-  @error_messages %{
-    not_found: "Task was not found.",
-    already_finished: "Cannott grab as task is already finished.",
-    already_assigned: "Cannot grab as task is already assigned."
-  }
+  def grab(%{assigns: %{current_user: %{id: volunteer_id}}} = conn, %{"task_id" => task_id}) do
+    grab_and_drop_common_handling(conn, :grab, task_id, volunteer_id)
```
ggpasqualino: what do you think of making `grab_and_drop_common_handling` just a response formatter? you can call the function that you want and pass the result to it instead of giving lots of params to be matched, for example
```elixir
Tasks.grab(task_id, volunteer_id) |> format_response(conn, "Task grabbed")
```

ghatighorias: done

PR #163, opened 2017-07-28
Added comments to attendances

laszlokardinal-dina:


web/templates/attendance/fill_course_attendance.html.eex
```diff
@@ -71,6 +71,56 @@
...
+                              data-attendance-id="<%= fh.id %>"
+                              class="mdl-button mdl-js-button mdl-button--icon show-attendance-modal"
+                            >
+                              <i
+                                class="
+                                  material-icons
+                                  attendance-comment__button
+                                  <%=
+                                    if fh.data.comment && String.length(fh.data.comment) > 0 do
+                                      'attendance-comment__button--set'
```
rhnonose: In Elixir, single-quoted words are char lists and double-quoted words are strings. You should use strings generally because char lists are usually used to interface with erlang. This might be a problem when using, for example, the `String` module.

web/templates/attendance/fill_course_attendance.html.eex
```diff
@@ -71,6 +71,56 @@
...
+                              data-attendance-id="<%= fh.id %>"
+                              class="mdl-button mdl-js-button mdl-button--icon show-attendance-modal"
+                            >
+                              <i
+                                class="
+                                  material-icons
+                                  attendance-comment__icon
+                                  attendance-comment__icon--fill
+                                  <%=
+                                    if fh.data.comment && String.length(fh.data.comment) > 0 do
```
ghatighorias: you can extract this condition to a function in view and then use "Changeset.get_field(changeset, key, default \\ nil)" instead of "fh.data"

ggpasqualino: @laszlokardinal-dina could you please check @ghatighorias comment?

PR #164, opened 2017-07-31
Adds about page for the program

ghatighorias:
**Note : this PR is depended on https://github.com/digitalnatives/course_planner/pull/170 to be merged first to the master, so this one becomes smaller

1. Adds new config parameters for describing a user program
2. Adds about page and populate it based on config params
3. Adds test to cover all cases


web/templates/about_client_program/show.html.eex
```diff
@@ -0,0 +1,86 @@
...
+        <div class="col-xs-6 col-sm-9 col-md-10 page-title">
+          About <%= @program_data.name %>
+        </div>
+      </div>
+
+      <div class="page">
+          <%= CoursePlanner.SharedView.card do %>
+            <%= CoursePlanner.SharedView.card_content vpadding: true do %>
+              <div class="row row--hspace row--vspace middle-xs">
+                <%= if not is_nil(@program_data.logo) do %>
```
rhnonose: `if` statement in Elixir checks for truthy values, meaning anything not `nil` or `false` is considered true, so you can check simply like this: `if @program_data.logo do`

ggpasqualino:
@ghatighorias could you check the conflicts, remove the duplicated migration?

web/router.ex
```diff
@@ -75,6 +75,7 @@ defmodule CoursePlanner.Router do
     end
 
     resources "/settings", SettingController, only: [:show, :edit, :update], singleton: true
+    resources "/aboutclientprogram", AboutClientProgramController, only: [:show], singleton: true
```
ggpasqualino: what if we name it just `about`?

ghatighorias: i wanted to , but don't we wanna have an about page for our program ?
( i mean Dina's course planner ) . cause if no then I will, i was just not sure :P

ggpasqualino: at this point one instance of course planner is for one program, so if Dina has a program it will have a normal about for that instance

ghatighorias: oki i change it to about and if there will be any other need we can decide later

ghatighorias: done

priv/repo/seeds.exs
```diff
@@ -43,28 +81,66 @@ CoursePlanner.Repo.delete_all CoursePlanner.User
       value: "some email",
       type: "string",
       editable: true,
-      visible: true
+      visible: false
```
ggpasqualino: missing the required field

ghatighorias: done

PR #169, opened 2017-08-01
Introduces categorized show and edit for system variables

ghatighorias:
1. Separates the system_variables to system and program and uses two different tables to list them in the show page
2. introduces two separate edit link for editing system and program related system variables separately
3. updates the router so edit can get an id to select editing between system or program system_variables
4. updates the current tests and introduces new ones

### DIFF
--- a/test/controllers/setting_controller_test.exs
+++ b/test/controllers/setting_controller_test.exs
@@ -13,20 +13,28 @@ defmodule CoursePlanner.SettingControllerTest do
     {:ok, conn: conn}
   end
 
+  @moduletag user_role: :coordinator
   describe "settings functionality for coordinator user" do
-    @tag user_role: :coordinator
     test "shows chosen resource only for coordinator", %{conn: conn} do
       conn = get conn, setting_path(conn, :show)
       assert html_response(conn, 200) =~ "Show settings"
     end
 
-    @tag user_role: :coordinator
-    test "renders form for editing for coordinator", %{conn: conn} do
-      conn = get conn, setting_path(conn, :edit)
+    test "renders form for editing system setting for coordinator", %{conn: conn} do
+      conn = get conn, setting_edit_path(conn, :edit, "system_settings")
       assert html_response(conn, 200) =~ "Edit settings"
     end
 
-    @tag user_role: :coordinator
+    test "renders form for editing program setting for coordinator", %{conn: conn} do
+      conn = get conn, setting_edit_path(conn, :edit, "program_settings")
+      assert html_response(conn, 200) =~ "Edit settings"
+    end
+
+    test "renders 404 if request is notsystem_settings nor program_settings", %{conn: conn} do
+      conn = get conn, setting_edit_path(conn, :edit, "random")
+      assert html_response(conn, 404)
+    end
+
     test "does not update chosen resource and renders errors when data is invalid for coordinator user", %{conn: conn} do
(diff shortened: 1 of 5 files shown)


web/controllers/setting_controller.ex
```diff
@@ -8,16 +8,39 @@ defmodule CoursePlanner.SettingController do
...
+        |> put_status(404)
+        |> render(CoursePlanner.ErrorView, "404.html")
+      _   ->
+        changeset =
+          filtered_system_variables
+          |> Enum.map(&SystemVariable.changeset/1)
+          |> Settings.wrap()
+
+        render(conn, "edit.html", changeset: changeset)
+    end
```
rhnonose: Can you trim this function a bit? It's too long.
Looks like you can "merge" both `case` statements and use a separate function to extract the `"setting_type"` or simply split it into two functions.

ghatighorias: I moved parts of the functionality out, check if you see it good enough

ggpasqualino:
@ghatighorias please check the conflicts

test/controllers/setting_controller_test.exs
```diff
@@ -116,20 +127,23 @@ defmodule CoursePlanner.SettingControllerTest do
     end
   end
 
+  @tag user_role: :volunteer
```
ggpasqualino: should this be `@module_tag` as well?

web/models/setting/settings.ex
```diff
@@ -37,6 +37,22 @@ defmodule CoursePlanner.Settings do
...
+  def filter_system_variables(system_variables, setting_type) do
+      case setting_type do
+        "system_settings"  -> {:ok, filter_non_program_systemvariables(system_variables)}
+        "program_settings" -> {:ok, filter_program_systemvariables(system_variables)}
+        _                  -> {:error, nil}
+      end
+  end
+
+  def filter_non_program_systemvariables(settings) do
+    Enum.filter(settings, &(not String.starts_with?(&1.key, "PROGRAM")))
```
ggpasqualino: please avoid negative conditions, what do you think of the following?
```elixir
Enum.reject(settings, &(String.starts_with?(&1.key, "PROGRAM")))
```

web/router.ex
```diff
@@ -74,7 +74,9 @@ defmodule CoursePlanner.Router do
       put "/update_fill", AttendanceController, :update_fill, as: :update_fill
     end
 
-    resources "/settings", SettingController, only: [:show, :edit, :update], singleton: true
+    resources "/settings", SettingController, only: [:show, :update], singleton: true do
```
ggpasqualino: if you don't change the router you can achieve a similar result using the following links
```elixir
setting_path(@conn, :edit, setting_type: "system") # /settings?setting_type=system
```
or
```elixir
setting_path(@conn, :edit, setting_type: "program") # /settings?setting_type=program
```
the controller can stay the same for this urls

the advantage would be having a standard path helper function

test/controllers/setting_controller_test.exs
ghatighorias: done

web/models/setting/settings.ex

web/router.ex
ghatighorias: nice trick, i'll take a note on this one

ghatighorias: done

PR #170, opened 2017-08-01
Adds new types to system variables

ghatighorias:
1. adds required field, and custom validation
2. makes the value field text for not be limited to 255 characters
3. introduce validation for string and url

### DIFF
--- a/priv/repo/migrations/20170801141406_change_system_variable_value_to_text_adds_required.exs
+++ b/priv/repo/migrations/20170801141406_change_system_variable_value_to_text_adds_required.exs
@@ -1,0 +1,10 @@
+defmodule CoursePlanner.Repo.Migrations.ChangeSystemVariableValueToTextAddsRequired do
+  use Ecto.Migration
+
+  def change do
+    alter table(:system_variables) do
+      modify :value, :text
+      add :required, :boolean
+    end
+  end
+end
--- a/priv/repo/migrations/20170801141406_change_system_variable_value_to_text_adds_required.exs
+++ b/priv/repo/migrations/20170801141406_change_system_variable_value_to_text_adds_required.exs
@@ -4,7 +4,7 @@ defmodule CoursePlanner.Repo.Migrations.ChangeSystemVariableValueToTextAddsRequi
   def change do
     alter table(:system_variables) do
       modify :value, :text
-      add :required, :boolean
+      add :required, :boolean, null: false, default: true
     end
   end
 end
--- a/test/models/system_variable_test.exs
+++ b/test/models/system_variable_test.exs
@@ -4,9 +4,9 @@ defmodule CoursePlanner.SystemVariableTest do
   alias CoursePlanner.SystemVariable
 
   @string_valid_attrs %{key: "sample key", value: "sample value", type: "string", visible: true, editable: true, required: true}
-  @integer_valid_attrs %{key: "sample key", value: "1", type: "integer", visible: true, editable: true}		 +  @text_valid_attrs %{key: "sample key", value: "sample value", type: "text", visible: true, editable: true, required: true}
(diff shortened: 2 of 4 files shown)


test/models/system_variable_test.exs
```diff
@@ -39,6 +41,70 @@ defmodule CoursePlanner.SystemVariableTest do
       changeset = SystemVariable.changeset(%SystemVariable{}, %{@string_valid_attrs | value: ""})
       refute changeset.valid?
     end
+
+    test "changeset fails when input is empty but field is not required" do
+      changeset = SystemVariable.changeset(%SystemVariable{}, %{@string_valid_attrs | value: "", required: false})
+      assert changeset.valid?
```
ggpasqualino: in the test description
> changeset fails

but it is valid

test/models/system_variable_test.exs
```diff
@@ -39,6 +41,70 @@ defmodule CoursePlanner.SystemVariableTest do
...
+    end
+
+    test "changeset fails when url has no scheme" do
+      changeset = SystemVariable.changeset(%SystemVariable{}, %{@url_valid_attrs | value: "www.sample.com"})
+      refute changeset.valid?
+    end
+
+    test "changeset fails when url has no host" do
+      changeset = SystemVariable.changeset(%SystemVariable{}, %{@url_valid_attrs | value: "http://"})
```
ggpasqualino: what is the usage of the url type? maybe we don't need to make the scheme required

web/models/setting/system_variable.ex
```diff
@@ -49,6 +50,16 @@ defmodule CoursePlanner.SystemVariable do
     |> validate_editable()
   end
 
+  def validate_by_record_required(changeset, target_params) do
```
ggpasqualino: should we call this function instead of `validate_required` in the update changeset?

test/models/system_variable_test.exs
ghatighorias: for the about page we wanted to add a link to a logo and to the website of the the program

web/models/setting/system_variable.ex
ghatighorias: hmmm "validate_required" is already a Ecto function and i didn't want to override it

test/models/system_variable_test.exs
ghatighorias: sorry this test should pass, cause the field is not required. i'll update the test name

web/models/setting/system_variable.ex
ggpasqualino: I don't mean you should rename this function, I mean that in the `changeset(struct, params, :update)` in line 48 we call the Ecto validate_required function. Should we change it to call this one instead?

test/models/system_variable_test.exs
ggpasqualino: can you change the error message from `Url should start with Http:// or Https://` to `Url should start with http:// or https://`?

web/models/setting/system_variable.ex
ghatighorias: ohhhh yes, sorry

ghatighorias: done

test/models/system_variable_test.exs
ghatighorias: done

priv/repo/migrations/20170801141406_change_system_variable_value_to_text_adds_required.exs
```diff
@@ -0,0 +1,10 @@
+defmodule CoursePlanner.Repo.Migrations.ChangeSystemVariableValueToTextAddsRequired do
+  use Ecto.Migration
+
+  def change do
+    alter table(:system_variables) do
+      modify :value, :text
+      add :required, :boolean
```
ggpasqualino: it would be nice to make this field not null, it is not very nice to have 3-state booleans, and have a default value as well, I think it can be true by default

ghatighorias: done

PR #172, opened 2017-08-02
creates summary page helpers

ghatighorias:
1. Makes term_factory and affected test independent from hardcoded date
2. Adds functionality for retriving terms and offered_course for users
3. adds a function to get the next_class

Note : the function for getting volunteer tasks is not in this pull request, it will be added after max_volunter pull request gets closed

### DIFF
--- a/test/models/summary_helper_test.exs
+++ b/test/models/summary_helper_test.exs
@@ -334,6 +334,53 @@ defmodule CoursePlanner.SummaryHelperTest do
     end
   end
 
+  describe "get_next_class function" do
+    test "when offered_course is an empty list" do
+      next_class = SummaryHelper.get_next_class([])
+
+      assert next_class == nil
+    end
+
+    test "when offered_course is not a list" do
+      next_class = SummaryHelper.get_next_class(nil)
+
+      assert next_class == nil
+    end
+
+    test "when offered_course have no class" do
+      offered_courses = insert_list(4, :offered_course) |> Repo.preload(:classes)
+      next_class = SummaryHelper.get_next_class(offered_courses)
+
+      assert next_class == nil
+    end
+
+    test "when offered_courses have classes on the same date with different hours" do
+      [offered_course1, offered_course2] = insert_list(2, :offered_course)
+      insert(:class, offered_course: offered_course1, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: 2))
+      class = insert(:class, offered_course: offered_course1, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: 1))
+      insert(:class, offered_course: offered_course2, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: 3))
+      insert(:class, offered_course: offered_course2, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: 4))
+
+      preload_offered_courses = preload_associations_for_offered_courses([offered_course1, offered_course2])
+      next_class = SummaryHelper.get_next_class(preload_offered_courses) |> Repo.preload([offered_course: [:course, :term]])
+
+      assert next_class == class
+    end
+
+    test "when offered_courses have classes on the same time with different date" do
+      [offered_course1, offered_course2] = insert_list(2, :offered_course)
+      insert(:class, offered_course: offered_course1, starting_at: Timex.now(), date: Timex.shift(Timex.now(), days: 2))
+      insert(:class, offered_course: offered_course1, starting_at: Timex.now(), date: Timex.shift(Timex.now(), days: 3))
+      class = insert(:class, offered_course: offered_course2, starting_at: Timex.now(), date: Timex.shift(Timex.now(), days: 1))
+      insert(:class, offered_course: offered_course2, starting_at: Timex.now(), date: Timex.shift(Timex.now(), days: 4))
+
+      preload_offered_courses = preload_associations_for_offered_courses([offered_course1, offered_course2])
+      next_class = SummaryHelper.get_next_class(preload_offered_courses) |> Repo.preload([offered_course: [:course, :term]])
+
+      assert next_class == class
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


web/models/summary_helper.ex
```diff
@@ -0,0 +1,80 @@
...
+      offered_courses
+      |> Enum.map(&(&1.term))
+      |> Enum.uniq()
+
+    %{terms: terms, offered_courses: offered_courses}
+  end
+  defp extract_data_from_offered_courses(_offered_courses), do: %{terms: [], offered_courses: []}
+
+  def get_next_class(offered_courses)
+    when is_list(offered_courses) and length(offered_courses) > 0 do
```
rhnonose: Is the `is_list` validation necessary? It's cleaner if you only validate empty list with pattern matching.

ghatighorias: yes, unless if you pass just one offered_course then it crashes

rhnonose: I mean, are you calling with one `offered_course` somewhere? Or is it "just in case"?

ghatighorias: it's totally a just in case :P
so the idea is that "get_term_offered_course_for_user" will return all the offered courses and using this function one can get the next classes. but I couldn't add it to the Map i return in "get_term_offered_course_for_user" cause coordinator and volunteer don't have classes

test/models/summary_helper_test.exs
```diff
@@ -0,0 +1,390 @@
...
+      student = insert(:student)
+      [term1, term2] = insert_list(2, :term)
+
+      insert_list(4, :offered_course, term: term1)
+      offered_course =
+        insert(:offered_course, term: term2, students: [student])
+        |> Repo.preload(:classes)
+
+      student_summary_data_response =
+        %{@empty_summary_helper_user_data_response | terms: [term2], offered_courses: [offered_course]}
```
ggpasqualino: why do you use this update map syntax? it seems unnecessary

web/models/summary_helper.ex
```diff
@@ -0,0 +1,80 @@
+defmodule CoursePlanner.SummaryHelper do
+  @moduledoc """
+    Provides helper functions for the summary page
+  """
+  use CoursePlanner.Web, :model
+
+  alias CoursePlanner.{Repo, Terms.Term, OfferedCourse}
+
+  def get_term_offered_course_for_user(user_id, user_role) do
```
ggpasqualino: what's the use case of this function? can you receive a `%User{}` instead of id and role separated?

test/models/summary_helper_test.exs
ghatighorias: so, you suggest of having two different assert ? or just simply create a new map ?

web/models/summary_helper.ex
ghatighorias: basically @laszlokardinal-dina want to call this from the controller. but we have the ID and role in the controller so I wrote it this way ( which i agree is not the best )

ghatighorias: it's gonna be one function for all users to get theiir personalized summary on term info

test/models/summary_helper_test.exs
ggpasqualino: create a new map

web/models/summary_helper.ex
ggpasqualino: in the controller there is the `%User{}` also, that's where we get the id from

ghatighorias: yesssss and we have the current_user as well. I am so stupid.
will fix this accordingly

ghatighorias: done

test/models/summary_helper_test.exs
ghatighorias: done

test/models/summary_helper_test.exs
```diff
@@ -0,0 +1,367 @@
...
+    insert_list(4, :offered_course, term: term1)
+    insert(:offered_course, term: term2, students: [student])
+
+    student_data = SummaryHelper.get_term_offered_course_for_user(%{student | role: "Unknown"})
+    assert student_data == @empty_summary_helper_user_data_response
+  end
+
+  describe "Student summary data" do
+    test "when she does not exist" do
+      student = build(:student, id: -1, role: "Students")
```
ggpasqualino: I don't think you need to give a role here, and there is an extra `s` in the end btw

ghatighorias: hmm that was left out

ghatighorias: done

web/models/summary_helper.ex
```diff
@@ -0,0 +1,80 @@
+defmodule CoursePlanner.SummaryHelper do
+  @moduledoc """
+    Provides helper functions for the summary page
+  """
+  use CoursePlanner.Web, :model
+
+  alias CoursePlanner.{Repo, Terms.Term, OfferedCourse}
+
+  def get_term_offered_course_for_user(%{id: user_id, role: role}) do
```
ggpasqualino: what do you think of adding a default parameter with the current time and then passing it down to the queries, like
```elixir
def get_term_offered_course_for_user(%{id: user_id, role: role}, current_time \\ Timex.now()) do
  case role do
    "Student" -> get_student_registered_data(user_id, current_time)
    "Teacher" -> get_teacher_registered_data(user_id, current_time)
    "Coordinator" -> get_all_terms_and_offered_courses(current_time)
    "Volunteer" -> get_all_terms_and_offered_courses(current_time)
    _           -> extract_data_from_offered_courses([])
  end
end
```
this makes the code easier to test because it doesn't depend on the real current time, and this also enable us to easily query the summary for yesterday or tomorrow if we want

ghatighorias: aha, haven't thought of "enable us to easily query the summary for yesterday or tomorrow", nice point. will do it

ghatighorias: done

PR #173, opened 2017-08-02
Notification summary

rhnonose:
Change notifications to be notified later and a button on user's index page to send all notifications.
Cron-jobs and notification frequency configuration will be added later.


lib/course_planner/users.ex
```diff
@@ -2,8 +2,9 @@ defmodule CoursePlanner.Users do
   @moduledoc """
     Handle all interactions with Users, create, list, fetch, edit, and delete
   """
-  alias CoursePlanner.{Repo, User, Notifier, Notifier.Notification}
+  alias CoursePlanner.{Repo, User, Notifications}
   alias Ecto.DateTime
+  @notifier Application.get_env(:course_planner, :notifier) || CoursePlanner.Notifier
```
ggpasqualino: please use `get_env/3` instead

test/test_helper.exs
```diff
@@ -3,3 +3,6 @@
 ExUnit.start
 
 Ecto.Adapters.SQL.Sandbox.mode(CoursePlanner.Repo, :manual)
+
+Code.compiler_options(ignore_module_conflict: true)
```
ggpasqualino: why is this necessary?

rhnonose: To supress a warning during test that the module was being loaded twice.
But now that I'm testing again, I don't have to load it here in the next line, so I'm removing it.

lib/course_planner/notifier.ex
```diff
@@ -43,8 +39,35 @@ defmodule CoursePlanner.Notifier do
...
+        delete_notifications(user)
+        {:noreply, state}
+      {:error, reason} ->
+        Logger.error("Email delivery failed: #{Kernel.inspect reason}")
+        {:noreply, [{:error, reason, email} | state]}
+    end
+  end
   def handle_cast(_, state), do: {:noreply, state}
 
+  defp delete_notifications(user) do
```
ggpasqualino: this implementation can delete notifications which were not sent yet if a notification is created between the notify_all casting and the delete notifications.
what do you think of mapping all the notification ids which are loaded in the user already and then delete by notification id?

rhnonose: You're right, I'll make the deletes with the IDs.

rhnonose:
@ggpasqualino I think we discussed tests about `notifier.ex` before about being hard hard to test with `Swoosh` because it sends messages to another process. I'm trying to isolate the logic into other modules in the other part of the notification changes.
I'll add tests for 'notification.ex`.

ggpasqualino:
@rhnonose yeah, I think extracting the body of the handle_cast's to plain module functions should help with that, in any case I think it is possible to call `CoursePlanner.Notifier.handle_cast({:notify_all, some_user}, [])` in the test, so the swoosh message will be sent to the test process

test/lib/course_planner/notifier_test.exs
```diff
@@ -0,0 +1,35 @@
...
+    |> Notifications.resource_path("/")
+    |> Notifications.to(user)
+
+    Notifier.handle_cast({:save_email, notification}, [])
+    saved_user = Repo.get(User, user.id) |> Repo.preload(:notifications)
+    [saved_notification] = saved_user.notifications
+    assert saved_notification.type == "user_modified"
+  end
+
+  test "send saved notifications" do
```
ggpasqualino: could you add a test about the deletion of the notifications as well?

web/controllers/user_controller.ex
```diff
@@ -68,4 +68,15 @@ defmodule CoursePlanner.UserController do
         |> redirect(to: user_path(conn, :index))
     end
   end
+
+  def notify(conn, _params) do
```
ggpasqualino: this is more like a safety net, but could you add a test for this also?

rhnonose: Hm, it falls in the process ownership problem with the sandbox. I'll see what I can do.

lib/course_planner/notifier.ex
```diff
@@ -32,7 +16,18 @@ defmodule CoursePlanner.Notifier do
...
+  end
+
+  @spec notify_all(User.t) :: GenServer.cast
+  def notify_all(%User{notifications: []} = _user), do: :nothing
+  def notify_all(%User{} = user) do
+    GenServer.cast(__MODULE__, {:notify_all, user})
+  end
+
+  @spec handle_cast({atom(), Notification.t} | atom(), any()) :: {:noreply, any()}
   def handle_cast({:send_email, notification}, state) do
```
ggpasqualino: do we need to keep this functionality?

rhnonose: My idea was to keep it in case we add any type of individual notification or even if we add an option to notify immediately instead of in a summary.

ggpasqualino: okay, can you add a test to it then? we left it behind the other time

rhnonose: Yeah, we tested only the e-mail building, I'll add.

rhnonose:
@ggpasqualino can you merge this branch with the decreased coverage?
I want to refactor to increase but I have a lot of code on top of this branch in `nonose/notifications_cron`.

PR #174, opened 2017-08-03
Uses shared templated for edit

ghatighorias:
1. Based on the type of the system variable loads different components for editing that system variable
2. uses a custom title for edit page to inform the user what type of system variables are being edited

### DIFF
--- a/web/views/setting_view.ex
+++ b/web/views/setting_view.ex
@@ -3,7 +3,7 @@ defmodule CoursePlanner.SettingView do
   use CoursePlanner.Web, :view
   alias CoursePlanner.SharedView
 
-  def get_form_component_from_type(form, field, type, label) do
+  def setting_input(form, field, type, label) do
     case type do
       "text"    -> SharedView.form_textarea(form, field, label: label)
       "string"  -> SharedView.form_text(form, field, label: label)
@@ -11,11 +11,7 @@ defmodule CoursePlanner.SettingView do
       "list"    -> SharedView.form_text(form, field, label: label)
       "integer" -> SharedView.form_text(form, field, label: label)
       "boolean" -> SharedView.form_select(form, field, ["True", "False"], label: label)
-      _         -> render_default_setting_input(form)
+      _         -> SharedView.form_text(form, field, label: label)
     end
-  end
-
-  defp render_default_setting_input(form) do
-    render "default_setting_input.html", form: form
   end
 end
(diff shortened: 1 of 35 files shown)


ghatighorias:
@laszlokardinal-dina I couldn't find any component in the shared folder to view a link. any idea how to do so ?

web/views/setting_view.ex
```diff
@@ -1,8 +1,21 @@
...
-    "Settings"
+  def get_form_component_from_type(form, field, type, label) do
+    case type do
+      "text"    -> SharedView.form_textarea(form, field, label: label)
+      "string"  -> SharedView.form_text(form, field, label: label)
+      "url"     -> SharedView.form_text(form, field, label: label)
+      "list"    -> SharedView.form_text(form, field, label: label)
+      "integer" -> SharedView.form_text(form, field, label: label)
+      "boolean" -> SharedView.form_select(form, field, ["True", "False"], label: label)
+      _         -> render_default_setting_input(form)
```
ggpasqualino: what do you think of using `SharedView.form_text(form, field, label: label)` as the default?

web/views/setting_view.ex
```diff
@@ -1,8 +1,21 @@
 defmodule CoursePlanner.SettingView do
   @moduledoc false
   use CoursePlanner.Web, :view
+  alias CoursePlanner.SharedView
 
-  def title do
-    "Settings"
+  def get_form_component_from_type(form, field, type, label) do
```
ggpasqualino: what do you think of naming this function `setting_input`?

ghatighorias: done

web/views/setting_view.ex
ghatighorias: done

web/views/setting_view.ex
```diff
@@ -1,8 +1,17 @@
 defmodule CoursePlanner.SettingView do
   @moduledoc false
   use CoursePlanner.Web, :view
+  alias CoursePlanner.SharedView
 
-  def title do
-    "Settings"
+  def setting_input(form, field, type, label) do
+    case type do
```
ggpasqualino: hat do you think of removing the cases which have the same behavior as the default? for example
```elixir
"text"    -> SharedView.form_textarea(form, field, label: label)
"boolean" -> SharedView.form_select(form, field, ["True", "False"], label: label)
_         -> SharedView.form_text(form, field, label: label)
```

ghatighorias: the same for me. I originally put them there in case we wanna show some different ways in future. but I can as well removed them for now

ghatighorias: done

PR #176, opened 2017-08-08
Removes the extra newline in term confirm message. center other messages

ghatighorias:
check comment section of this card for the description of the bug https://app.nostromo.io/work/course-planner/update-deletion-warning-messeges?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjkzNDEwLGNhcmRfdGl0bGU6dXBkYXRlLWRlbGV0aW9uLXdhcm5pbmctbWVzc2VnZXN9


ghatighorias:
Anyone one knows a good solution to make sure "are you sure?" is centered ? now i did it manually

ggpasqualino:
@ghatighorias I think there is no easy way to style the confirmation dialog.

Also, I think it would be better not to try to style it at all. 
Browsers can show the message differently and break lines in different positions.

ghatighorias:
then shall i remove the spaces before "are you sure?"

ggpasqualino:
@ghatighorias yeah, also make the explanation text one line

PR #177, opened 2017-08-08
Notification cron

rhnonose:
* Add notified ant notification frequency attributes to user
* Add cron to send notification to all notifiable users every day 6PM


priv/repo/migrations/20170803180536_add_notification_configuration.exs
```diff
@@ -0,0 +1,17 @@
+defmodule CoursePlanner.Repo.Migrations.AddNotificationConfiguration do
+  use Ecto.Migration
+
+  def up do
+    alter table(:users) do
+      add :notified, :date
```
ggpasqualino: please rename this field to `notified_at`, it is the convention to use `*_at` for the timestamp of an action, for example `inserted_at`, or `updated_at`

coveralls.json
```diff
@@ -3,7 +3,8 @@
     "minimum_coverage": 87
   },
   "skip_files": [
-    "test/support"
+    "test/support",
+    "lib/course_planner/notifier.ex"
```
ggpasqualino: please remove this from the exceptions, we should know that this file is not being covered when we see the coveralls result, if we can't/won't cover this will be an informed decision

priv/repo/migrations/20170803180536_add_notification_configuration.exs
```diff
@@ -0,0 +1,17 @@
...
+      add :notification_frequency_days, :integer, null: false, default: 1
```
ggpasqualino: I am not sure about `frequency` in the name, when i see this name I think of `1 message/day`, or `2 message/day`

lib/course_planner/users.ex
```diff
@@ -44,6 +47,19 @@ defmodule CoursePlanner.Users do
     |> @notifier.notify_later()
   end
 
+  def update_notifications(user) do
+    Multi.new()
+    |> delete_notifications(user.notifications)
```
ggpasqualino: what do you think of leaving lines 52 and 53 in the same level of abstraction? It may be something like
```elixir
Multi.new()
  |> Multi.delete_all(Notification, notifications_to_delete(params))
  |> Multi.update(User, Changeset.change(user, notified: Ecto.Date.utc()))
  |> Repo.transaction()
```
or
```elixir
Multi.new()
  |> delete_notifications(user.notifications)
  |> update_last_notified(params)
  |> Repo.transaction()
```

priv/repo/migrations/20170803180536_add_notification_configuration.exs
ggpasqualino: also this should probably be a timestamp not a date

web/models/coherence/user.ex
```diff
@@ -21,6 +22,8 @@ defmodule CoursePlanner.User do
     field :comments, :string
     field :role, UserRole
     field :participation_type, ParticipationType
+    field :notified, Ecto.Date
+    field :notification_frequency_days, :integer
```
ggpasqualino: this field should have a minimum/maximum validation, probably from 1 to 7

priv/repo/migrations/20170803180536_add_notification_configuration.exs
ghatighorias: hmm is it number of message per day, or message in every x day ?

ggpasqualino: *frequency* is the number of times an event happens in a given time, for example 3 notifications/day
*period* is the amount of time between two events, for example 1 day/notification

this is what comes in my mind when I see this naming, there may be other definitions which I am not aware of...

lib/course_planner/scheduler.ex
```diff
@@ -0,0 +1,5 @@
+defmodule CoursePlanner.Scheduler do
```
ghatighorias: we already have a module name ScheduleController and ScheduleView. what do you think of making this one a bit more explicit. something like "email_scheduler" or etc

config/config.exs
```diff
@@ -34,4 +34,9 @@ config :canary,
   repo: CoursePlanner.Repo,
   unauthorized_handler: {CoursePlanner.Helper, :handle_unauthorized}
 
+config :course_planner, CoursePlanner.Scheduler,
+  jobs: [
+    {"0 18 * * *", {CoursePlanner.Notifications, :send_all_notifications, []}}
```
ghatighorias: i am not sure how does this work. in case the server go to sleep and wakes up again does this scheduler runs immediately ? ( if so everything is alrighty)

coveralls.json
rhnonose: Can we mark as not relevant in some way? I isolated the hardly testable part of the code to this file and I believe it shouldn't be treated as "missing coverage" when it's not coveralble (or at least it's hard to).

ggpasqualino: but this is relevant, as I said before
>  if we can't/won't cover this will be an informed decision

config/config.exs
rhnonose: There's a problem in heroku free tier with crons. For now I didn't find a good solution, but the job does not run if the instance is sleeping.

web/models/coherence/user.ex
ghatighorias: @ggpasqualino @rhnonose  should it be both database and application level or just application level would suffice ?

lib/course_planner/notifications.ex
```diff
@@ -16,4 +17,18 @@ defmodule CoursePlanner.Notifications do
   def to(%Notification{} = notification, %User{} = user),
     do: %{notification | user: user}
 
+  def send_all_notifications do
+    Timex.today()
```
ghatighorias: Do we need to have one system setting like ENABLE_NOTIFICAITON that completely stop all notifications ?

priv/repo/migrations/20170803180536_add_notification_configuration.exs
ghatighorias: with having a notification_frequency per user we don't need the system_variable "NOTIFICATION_FREQUENCY" could you delete it from the seed file ?

lib/course_planner/notifications.ex
rhnonose: Not sure, why would you need that?

web/models/coherence/user.ex
rhnonose: I believe application level is enough, I'm putting also a select in the templates.

lib/course_planner/notifications.ex
ghatighorias: i mean by limiting the notification between 1-7 days it is good to give the coordinator the power to stop all the notifications.
e.g. they first fill everything in the application and then activate the notification

ghatighorias: @ggpasqualino any idea on this ?

rhnonose: Yeah, got it. Card created.

ggpasqualino: it seems like a valid use case...

PR #185, opened 2017-08-10
Disable notifications through settings.

rhnonose:


priv/repo/seeds.exs
```diff
@@ -133,3 +133,15 @@ CoursePlanner.Repo.delete_all CoursePlanner.User
...
+
+%CoursePlanner.SystemVariable{}
+|> CoursePlanner.SystemVariable.changeset(
+    %{
+      key: "DISABLE_NOTIFICATION",
+      value: "true",
+      type: "boolean",
+      editable: true,
+      visible: true,
+      required: false
```
ghatighorias: i think this should be required: true
it means it cannot be null

rhnonose: Isn't it enough to call `Settings.get_value("DISABLE_NOTIFICATION", false)` with the default value?

ghatighorias: that is alright. Required field controls the update action. it allows the user to keep it as nil.

ghatighorias: however I am not sure if ecto allows that for boolean but i assume it would.
check the system_variable.ex for the detail function of :required field

lib/course_planner/notifications.ex
```diff
@@ -18,10 +18,11 @@ defmodule CoursePlanner.Notifications do
     do: %{notification | user: user}
 
   def send_all_notifications do
-    IO.puts "cron call"
-    Timex.today()
-    |> get_notifiable_users()
-    |> Enum.each(&Notifier.notify_all/1)
+    unless Settings.get_value("DISABLE_NOTIFICATION", false) do
```
ggpasqualino: can you change this condition to a positive condition? for example
```elixir
if Settings.get_value("NOTIFICATION_ENABLED", true) do
...
end
```

PR #198, opened 2017-08-28
Applied the logo from settings

laszlokardinal-dina:


rhnonose:
@ghatighorias changed the empty URL return to `nil`, so this should be updated.

laszlokardinal-dina:
done

PR #199, opened 2017-08-28
Laszlo.kardinal/different ui for roles

laszlokardinal-dina:


lib/course_planner_web/controllers/course_controller.ex
```diff
@@ -30,17 +30,6 @@ defmodule CoursePlannerWeb.CourseController do
...
-    case Repo.get(Course, id) do
-      nil ->
-        conn
-        |> put_status(404)
-        |> render(CoursePlannerWeb.ErrorView, "404.html")
-      course ->
-        render(conn, "show.html", course: course)
-    end
-  end
-
```
rhnonose: Why is the `show` action being removed?

laszlokardinal-dina: Because it has only 2 fields that is already shown on the index list

PR #203, opened 2017-08-29
bug fix : no function clause matching in String.Chars.URI.to_string/1

ghatighorias:
Settings.get_value function should return nil in case the value saved for the setting is nil so it can use the default value

this is especially important  for url type setting with empty string. we have already save it as nil and now we should consider get_value returning the default value in case system_variable.value is equal to nil

### DIFF
--- a/lib/course_planner_web/models/setting/settings.ex
+++ b/lib/course_planner_web/models/setting/settings.ex
@@ -29,9 +29,10 @@ defmodule CoursePlanner.Settings do
     system_variable = Repo.get_by(SystemVariable, key: name)
 
     {:ok, parsed_value} =
-      case system_variable do
-        nil -> {:ok, default}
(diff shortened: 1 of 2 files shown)


ggpasqualino:
this doesn't solve the problem completely, when the value is an empty string the problem still occurs

ghatighorias:
i had a previous PR that for url when the string is empty it returns nil
check system_variabl.ex line 111

ggpasqualino:
> i had a previous PR that for url when the string is empty it returns nil
exactly, it returns nil, not the default value, as far as I understood it should return the default value as well

ggpasqualino:
@ghatighorias so I checked it out and actually the empty string will only happen when we run the seed, otherwise when we save an empty value it will be nil.

So can you change the in the seeds the value from `""` to `nil`, please?

PR #204, opened 2017-08-29
Move/Rename modules according to phoenix contexts

rhnonose:
- Moved models out of `web` folder
- Grouped modules inside context folders
- Renamed modules' namespace according to context
There's a lot of changes, it's maybe easier to review the structure instead of the diffs.


ggpasqualino:
@ghatighorias can please review rodrigo's pr?

PR #205, opened 2017-08-29
add notification for missing attendances

ghatighorias:


test/support/test_notifier.ex
```diff
@@ -1,5 +1,10 @@
 defmodule CoursePlanner.TestNotifier do
+
+  alias CoursePlanner.{Notification, Notifier.Server}
+
   def notify_user(_), do: :ok
-  def notify_later(_), do: :ok
+  def notify_later(%Notification{} = notification) do
+    Server.handle_cast({:save_email, notification}, "test")
+  end
```
rhnonose: This `TestNotifier` module is a fake module to supress errors during test, so you're not supposed to call the "real" notifier here.

lib/course_planner/offered_courses.ex
```diff
@@ -74,4 +76,29 @@ defmodule CoursePlanner.OfferedCourses do
...
+  def notify_missing_attendances do
+    create_pending_attendance_notification_map()
+    |> Enum.each(fn(email_data) ->
+         Notifications.new()
+         |> Notifications.type(email_data.type)
+         |> Notifications.resource_path(email_data.path)
+         |> Notifications.to(email_data.user)
+         |> @notifier.notify_later()
+       end)
+  end
```
rhnonose: I'm not sure if this functions should belong in this module, maybe inside `UserEmail`?

test/support/test_notifier.ex
ghatighorias: then who am i suppose to test whether the record is created or not ?

lib/course_planner/offered_courses.ex
ghatighorias: i merely followed the rest of the code

test/support/test_notifier.ex
ghatighorias: I think i disagree, the notification should be tested black boxed. i cannot during my test rely on the fact that it works

rhnonose: I think there was a discussion before when I raised this issue, @ggpasqualino suggested to test the notifier separately (which is already being done). You separate the part that test the workflow of creating the notification and the part that persists.

lib/course_planner/offered_courses.ex
rhnonose: Ah, my bad, it's doing the way you're doing. I guess I wanted to refactor it but it's not done yet.

ghatighorias: Although I kinda agree with you, i was already extracting this repetitive part back to notificatins

test/support/test_notifier.ex
ghatighorias: but I am not testing the emails, I am only testing the function i wrote.
i agree with testing notifications separately cause they are written separately, but for my case i am testing "notify_missing_attendances" function

rhnonose: Yes, in this case you should test the "query and build the missing attendances" and "persist the created notifications" separately.
But if the way you wrote it is not raising errors, I guess it's fine. The fake module was being used because, when testing controller calls, it was raising errors because the repo process was not owned by the calling process.

ggpasqualino: I think the main issue is when we actually notify, the gen sever talks to swoosh

lib/course_planner_web/templates/email/attendance_missing.html.eex
```diff
@@ -0,0 +1,5 @@
+<div>
```
ggpasqualino: Did you check how this look like? I am not sure you have those variable available, for example `@teacher_names` and `@offered_course_name`

ghatighorias: hmm god point. no i haven't :P
this PR was on WIP and it seems i missed this part of it completely.
sorry will fix it now

lib/course_planner/notifications.ex
```diff
@@ -17,13 +17,25 @@ defmodule CoursePlanner.Notifications do
...
   def wake_up(now \\ DateTime.utc_now) do
     executed_at = Settings.get_value("NOTIFICATION_JOB_EXECUTED_AT", now)
     if Timex.diff(now, executed_at, :days) >= 1 do
+      build_all_notifications(now)
       send_all_notifications(now)
     end
   end
 
+  def build_all_notifications(now \\ DateTime.utc_now) do
+    if Settings.get_value("ENABLE_NOTIFICATION", true) do
```
ggpasqualino: we check this when we send notifications already, don't you want to create the notification in the db at all?

ghatighorias: ohh that's true I should create them regardless of the "ENABLE_NOTIFICATION"

lib/course_planner_web/templates/email/attendance_missing.html.eex
ghatighorias: Done. there is now a test for this

lib/course_planner/notifications.ex
ghatighorias: done

PR #211, opened 2017-09-06
Fixes the formatting of the class time

ghatighorias:
Since starting_at and finishes_at are Ecto.Time format it cannot be directly formatted by Timex.

Timex first need to convert them to naive_datetime and that need a full DateTime.

To solve this issue we need to use the class date to convert class times to DateTime and then format them as we wish


rhnonose:
I think this is due to ecto 2.2 change that I made, didn't realise it was breaking the formatting.
Good catch.

ghatighorias:
I noticed it in yesterday's demo :P :D

ggpasqualino:
> I think this is due to ecto 2.2 change that I made, didn't realise it was breaking the formatting. Good catch.

@rhnonose when was this done?

PR #212, opened 2017-09-06
Fixes the app so coherence sends it's emails

ghatighorias:
Root cause :
in tag 0.4.0 Coherence.Config function "mailer?" is defined with hard coded module names for the example demo
` def mailer? do
    !!Application.get_env(:coherence, CoherenceDemo.Coherence.Mailer)
  end`

this problem is solved in tag 0.5.0 with the following implementation
`  def mailer? do
    !!Application.get_env(:coherence, Module.concat(web_module(), Coherence.Mailer))
  end`

but the tag had an issue with the latest 1.3 phoenix 1.3 and we had to revert back to tag 0.4.0

additionally in tag 0.4.0 in many places coherence mailer is located using "Coherence.Config.module" function with does not return "CoursePlannerWeb" but "CoursePlanner"

Solution : 
 the suggestion in this PR is rather a temporary fix and should not be considered as a fix forever


rhnonose:
That's a weird cause.
It's also weird that it gives no error/warning/info either.

ghatighorias:
it does give an error, it's just ignored in our program.
`send_user_email` function returns `{:error, :no_mailer}` but basically we do not check the return value of the function anywhere

ggpasqualino:
this fix should not be done here

- please fork the library and fix it there, than we will use the fork
- also open an issue on the main repo of the library for them to make a patch release of 0.4.X

ghatighorias:
@ggpasqualino 0.4.0 is a tag and not a branch. it would be more complicated to fix it

ghatighorias:
this PR will be closed and https://github.com/digitalnatives/course_planner/pull/219 will replace it by upgrading coherence to 0.5.0

PR #213, opened 2017-09-06
Ensures the creation of the default admin if there is none

ghatighorias:
check the link for the scenario and the description
 https://app.nostromo.io/work/course-planner/all-admin-pending-dilemma?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjk1OTU4LGNhcmRfdGl0bGU6YWxsLWFkbWluLXBlbmRpbmctZGlsZW1tYX0=


priv/repo/seeds.exs
```diff
@@ -1,21 +1,21 @@
...
-        email: "testuser@example.com",
-        password: "secret",
-        password_confirmation: "secret",
-        role: "Coordinator"
-      },
-      :seed)
-  |> Repo.insert!
+default_coordinator = Repo.get_by(User, email: "admin@courseplanner.com")
+if default_coordinator do
+  Repo.delete!(default_coordinator)
```
ggpasqualino: why delete it?

ghatighorias: I don't know, I thought maybe they change the password or something, so I just thought of having some resetting for this user, to ensure it will get created always with the default password.

what do you think about it ?

ggpasqualino:
this task is not necessary anymore, the email sending was fixed and coordinators can't delete themselves anymore as well, which prevents the bad state this PR was trying to work around

PR #215, opened 2017-09-06
introduces bulk creation for users and as a generic approach

ghatighorias:
under this PR the followings are implemented

1. CSV parser = parse csv string according to the requested column with the following functionalities

1.1. validate inputs against reuested number of columns
1.2. optional trimming of space before or after of the elements in csv after splitting them
1.3. returning the data in lists, each of which have the length of the requested number columns
 

2. Bulk controller = that provides a generic solution for creation of any type of entity
3. A specific implementation for added for creating users to the Bulk controller
4. bulk user creation button is added to "all users" page
5. tests for points 1,2 and 3


lib/course_planner/accounts/user.ex
```diff
@@ -45,6 +45,12 @@ defmodule CoursePlanner.Accounts.User do
       greater_than_or_equal_to: 1, less_than_or_equal_to: 7)
   end
 
+  def changeset(model, params, :bulk) do
+    model
+    |> changeset(params)
+    |> validate_required([:role])
```
rhnonose: Why is `role` the only required field in here? Shouldn't `email` be here as well?

lib/course_planner/csv_parser.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  defp trim_splitted_csv(splited_csv) do
+    Enum.map(splited_csv, &(String.trim(&1)))
+  end
+
+  defp csv_splitter(csv_data) do
+    String.split(csv_data, @csv_seperator)
+  end
+
+  defp validate_column_count(trimmed_splitted_csv, column_count) do
```
rhnonose: Now this validation looks weird to me.
So you're ignoring the line-breaks and validating the sum of all record's columns?
I'm not sure if this is a good validation. I think every record should end in a line break and the validation should have a per row granularity.

lib/course_planner/accounts/user.ex
ghatighorias: because email is already is a required field in "changeset(params)" but not role

rhnonose: Sorry, I forgot my eyes somewhere :D

lib/course_planner/csv_parser.ex
ghatighorias: so what if I remove the new lines here and then validate?

rhnonose: I think you should split first by line separator and validate them individually.
Let's say there's 10 lines but only one line is invalid. Should you insert the valid 9 records?
If not, you should specifically warn that line x is invalid. Think about inserting a massive CSV and then you get a generic "your CSV is invalid" warning.
This also removes the necessity to chunk the records and avoid false-positives: e.g. there's 5 lines of with 4 records each, it's gonna consider valid in terms of number of records if you don't validate the line break.

ghatighorias: I'll update the generic error too.
but I think it's better to insert only if everything is okay

ghatighorias: Done

ghatighorias:
@rhnonose the permission is automatically restricted by the following two lines in the permissions
this line allowed coordinator all the rights `  def can?(%User{role: "Coordinator"}, _action, _controller), do: true`
this line take all the rights from the other
`  def can?(_user, _action, _controller), do: false`

lib/course_planner/csv_parser.ex
```diff
@@ -0,0 +1,54 @@
+defmodule CoursePlanner.CsvParser do
```
ggpasqualino: why did you implement a CSV parser?
CSV parsing is much more complex than this, you should probably look for a library

lib/course_planner/accounts/user.ex
ggpasqualino: 🤔  role should probably be required in the default changeset

lib/course_planner_web/templates/bulk/form.html.eex
```diff
@@ -0,0 +1,51 @@
...
+              </div>
+            </div>
+          </div>
+
+          <%= hidden_input(f, :title, value: @title) %>
+          <%= hidden_input(f, :target, value: @target) %>
+
+          <%= CoursePlannerWeb.SharedView.card @title do %>
+            <%= CoursePlannerWeb.SharedView.card_content do %>
+              <%= CoursePlannerWeb.SharedView.form_textarea f, :csv_data %>
```
ggpasqualino: this is a bit odd way to work with csv's, usually people upload the csv file because is much easier to handle it in excel or something like that

ghatighorias: well we don't have any custom uploading component for it and i thought i just make it simple

lib/course_planner/csv_parser.ex
ghatighorias: I wanted to write something simple with column validation within the parser, so in this sense this is rather interpret csv than parsing it.
but why do you say it is much more complex ? like what part of it should be different or what scenarios it does not cover ?

ghatighorias:
replaced with https://github.com/digitalnatives/course_planner/pull/224

PR #216, opened 2017-09-06
Fixes the error when using the password reset link

ghatighorias:


ggpasqualino:
I am not sure this is the right approach.

The problem is that for some reason the password controller is using the wrong view (CoursePlannerWeb.Coherence.SessionView) when it should be using CoursePlannerWeb.Coherence.PasswordView

ghatighorias:
to be honest I couldn't find out exactly why such a thing happen, I create it this as a patch that can solve our problem at least in short term

ggpasqualino:
Closed in favor of https://github.com/digitalnatives/course_planner/pull/221 as I mentioned before the problem was not a missing template

>The problem is that for some reason the password controller is using the wrong view (CoursePlannerWeb.Coherence.SessionView) when it should be using CoursePlannerWeb.Coherence.PasswordView

PR #218, opened 2017-09-07
Adds more test for missing attendance notification creation

ghatighorias:
I adding more test for creation of notification for the missing attendances for teachers


ghatighorias:
@ggpasqualino I have added some more test as well, but I still can't reproduce the bug. can you have a look at the tests i wrote for the missing attendance notification creation functionality  and tell if I am missing something

ghatighorias:
I think as for a final check i can test "build_all_notifications" function from "CoursePlanner.Notifications" module as well

ggpasqualino:
this is almost the same as the test `when there is one offered_course with one teacher have missing attendance` but notice that I change the `updated_at` value of the teacher when I ask for their notification. This attribute should not matter at all for this process (actually none of them should matter besides the id), but it breaks.

```elixir
    test "bah" do
      students = insert_list(3, :student)
      teacher = insert(:teacher)
      [class1, class2] = insert_list(2, :class, date: Timex.shift(Timex.now(), days: -2))

      Enum.each(students, fn(student) ->
       insert(:attendance, student: student, class: class1, attendance_type: "Not filled")
       insert(:attendance, student: student, class: class2, attendance_type: "Not filled")
      end)

      offered_course = insert(:offered_course, classes: [class1, class2], students: students, teachers: [teacher])

      OfferedCourses.create_missing_attendance_notifications([%{teacher | updated_at: NaiveDateTime.utc_now}])
      [notification] =
        Notification
        |> Repo.all()
        |> Repo.preload(:user)
        |> Enum.sort(&(&1.user.id <= &2.user.id))

      assert notification.user == teacher
      assert notification.type == "attendance_missing"
      assert notification.resource_path == Attendances.get_offered_course_fill_attendance_path(offered_course.id)
    end
```

ghatighorias:
Thanks a lot. tonight I take a look into it

test/lib/course_planner/courses/offered_courses_test.exs
```diff
@@ -336,4 +406,28 @@ defmodule CoursePlanner.OfferedCoursesTest do
 
     end
   end
+
+  test "bah" do
```
ggpasqualino: I think you don't need to add this test after the change on line 255, specially with this name 😅

ggpasqualino: I think you don't need to add this test after the change on line 255, specially with this name

ghatighorias: ohh darn, i forgot it there . damn

ghatighorias: done

lib/course_planner/courses/offered_courses.ex
```diff
@@ -79,19 +79,19 @@ defmodule CoursePlanner.Courses.OfferedCourses do
   end
 
   def create_pending_attendance_notification_map(notifiable_users) do
```
ggpasqualino: I thought about an alternative solution using for-comprehension for this, which seems more understandable
```elixir

  def create_pending_attendance_notification_map(notifiable_users) do
    notifiable_ids = Enum.map(notifiable_users, &(&1.id))
    for oc <- with_pending_attendances(), t <- oc.teachers, t.id in notifiable_ids do
      %{
        user: t,
        type: :attendance_missing,
        path: Attendances.get_offered_course_fill_attendance_path(oc.id),
        data: %{offered_course_name: "#{oc.term.name}-#{oc.course.name}"}
      }
    end
  end
```
what do you think?

ggpasqualino: if you like the solution you may use better names though 😅

ghatighorias: looks pretty cool

ghatighorias: i think to be honest naming is better this way. when I tried to replace t with teacher and oc with offered_course it was a bit confusing already for me to understand it

ghatighorias: done. replaced

PR #219, opened 2017-09-08
Upgrade to coherence tag 0.5.0

ghatighorias:


ggpasqualino:
there is a function to get the flash message in the controller tests `get_flash`, so you should be able to do something like
```elixir
assert get_flash(conn, "error") == "Coordinator was not found."
```
instead of 
```elixir
assert conn.private.plug_session == %{"phoenix_flash" => %{"error" => "Coordinator was not found."}}
```

ghatighorias:
get_flash replacement is done

PR #220, opened 2017-09-08
Resend user created notification.

rhnonose:


rhnonose:
Pending #219 changes (tested successfully).

lib/course_planner_web/controllers/user_controller.ex
```diff
@@ -75,4 +77,19 @@ defmodule CoursePlannerWeb.UserController do
     |> put_flash(:info, "Users notified successfully.")
     |> redirect(to: user_path(conn, :index))
   end
+
+  def resend_email(conn, %{"id" => id}) do
+    token = ControllerHelpers.random_string 48
+    url = Helpers.password_url(conn, :edit, token)
```
ghatighorias: is it correct ?
you are generating a new link for the user to acess the reset password while there is already one existing and saved to him. so shouldn't you first update the user token with the newly generated one then tun send_user_email ?

rhnonose: You might be right, I'll test that case.

ggpasqualino: @ghatighorias is right, this will generate an invalid url

lib/course_planner_web/controllers/user_controller.ex
```diff
@@ -75,4 +77,19 @@ defmodule CoursePlannerWeb.UserController do
...
+    case Users.get(id) do
+      {:ok, user} -> ControllerHelpers.send_user_email(:welcome, user, url)
```
ggpasqualino: can you put `ControllerHelpers.send_user_email(:welcome, user, url)` in the next line? this is a bit confusing

lib/course_planner_web/router.ex
```diff
@@ -55,6 +55,7 @@ defmodule CoursePlannerWeb.Router do
 
     resources "/users", UserController, except: [:create, :new]
     post "/notify", UserController, :notify
+    put "/resend_email/:id", UserController, :resend_email
```
ggpasqualino: why `put`?

rhnonose: Not sure, I guess because it had the user's ID at hand.

PR #221, opened 2017-09-11
Use the assigned view

ggpasqualino:
It was not possible to open the `Forgot your password?` and the reset
password link from the email because we were trying to render always
from the SessionView.


ghatighorias:
nice catch

PR #222, opened 2017-09-11
Remove "NOTIFICATION_JOB_EXECUTED_AT" from the edit page

ghatighorias:
Currently "NOTIFICATION_JOB_EXECUTED_AT" is shown on the edit page and with this PR it will be removed from the edit page

### DIFF
--- a/lib/course_planner_web/controllers/setting_controller.ex
+++ b/lib/course_planner_web/controllers/setting_controller.ex
@@ -63,7 +63,7 @@ defmodule CoursePlannerWeb.SettingController do
           changesets
           |> Enum.sort_by(&(Changeset.get_field(&1, :key)), &<=/2)
           |> Settings.wrap()
-          |> Map.put(:action, :update)
+          |> Map.put(:action, :user_update)
         render(conn, "edit.html", changeset: changeset, title: title)
     end
   end
--- a/lib/course_planner_web/controllers/setting_controller.ex
+++ b/lib/course_planner_web/controllers/setting_controller.ex
@@ -63,7 +63,7 @@ defmodule CoursePlannerWeb.SettingController do
           changesets
           |> Enum.sort_by(&(Changeset.get_field(&1, :key)), &<=/2)
           |> Settings.wrap()
-          |> Map.put(:action, :user_update)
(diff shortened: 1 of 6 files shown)


ggpasqualino:
this PR breaks `CoursePlanner.Notifications.update_executed_at/1`

ghatighorias:
yes that's why i changed it back to WIP. I am trying to find a good solution for this

rhnonose:
Doesn't the `visible` tag supposed to hide in the form?

ghatighorias:
Edit form reads all the editable tags and not the visible ones cause there might be ones that are visible but not editable. speaking of combination "visible = false, editable = true" should not exist for any system variable. only "visible = true, editable = true" and "visible = true, editable = false" and "visible = false, editable = false" should exist.

rhnonose:
Well, I guess this is the case for `visible = false, editable = true`.

ghatighorias:
I think I have found the problem
the function "def changeset(struct, params, :update)" in system_variable.ex is only for when user wanna update a variable and it will check the :editable field. this function is used in notifications for updating "NOTIFICATION_JOB_EXECUTED_AT" and instead  "def changeset(struct, params)" should be used.

to fix this I change `def changeset(struct, params, :update)` to `def changeset(struct, params, :user_update)` so it becomes more clear for us.

lib/course_planner_web/controllers/setting_controller.ex
```diff
@@ -63,7 +63,7 @@ defmodule CoursePlannerWeb.SettingController do
           changesets
           |> Enum.sort_by(&(Changeset.get_field(&1, :key)), &<=/2)
           |> Settings.wrap()
-          |> Map.put(:action, :update)
```
ggpasqualino: this action is not related, this is an [Ecto.Changeset action](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-changeset-actions)

ggpasqualino: this action is not related, this is an [Ecto.Changeset action](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-changeset-actions)

PR #223, opened 2017-09-12
Fixes shaky tests

ghatighorias:


test/lib/course_planner/classes/summaries_test.exs
```diff
@@ -54,7 +54,11 @@ defmodule CoursePlanner.SummariesTest do
         |> Repo.preload(:classes)
 
       student_data = Summaries.get_term_offered_course_for_user(student)
-      assert student_data == %{terms: [term2], offered_courses: offered_courses}
+      expected_data_offered_courses = Enum.sort(offered_courses, &(&1.id >= &2.id))
```
rhnonose: You can extract this to something like `sort_by_id(offered_courses)`, you use it a lot of times.

PR #224, opened 2017-09-12
bulk user insert with file upload

ghatighorias:
This solution uses CSV library and file upload as suggested in https://github.com/digitalnatives/course_planner/pull/215 PR


lib/course_planner_web/models/bulk_helper.ex
```diff
@@ -0,0 +1,78 @@
...
+    |> CSV.decode
+    |> Enum.reduce_while({:ok, []}, fn(parsed_row, {_out_result, out_value}) ->
+         case parsed_row do
+           {:ok, value} -> {:cont, {:ok, [value | out_value]}}
+           {:error, value} -> {:halt, {:error, value}}
+         end
+       end)
+  end
+
+  def validate_column_count({:error, parsed_params}, _column_count), do: {:error, parsed_params}
```
ggpasqualino: you don't need to do this validation, the library will give you an error if one of the lines is bigger or smaller than the others, this includes empty lines

ghatighorias: so is it enough if I check the first line then ? to see if it has specifically 5 columns ?

lib/course_planner_web/models/bulk_helper.ex
```diff
@@ -0,0 +1,78 @@
...
+      {:error, :empty} ->
+        {:error, "parsing_csv", "Input can not be empty.", ""}
+      {:error, message} ->
+        {:error, "parsing_csv", message, ""}
+    end
+  end
+
+  defp csv_parsing(csv_file_stream) do
+    csv_file_stream
+    |> CSV.decode
```
ggpasqualino: I think it would be more interesting to call `CSV.decode(strip_fields: true, headers: %w(name, family_name, nickname, email, role)a)`

this will tell the library the number of columns you are expecting already, besides that it creates already maps for each line, and it will strip the columns

lib/course_planner/accounts/user.ex
```diff
@@ -46,6 +46,12 @@ defmodule CoursePlanner.Accounts.User do
       greater_than_or_equal_to: 1, less_than_or_equal_to: 7)
   end
 
+  def changeset(model, params, :bulk) do
+    model
+    |> changeset(params)
+    |> validate_required([:role])
```
ggpasqualino: I think we should add this validation to the default changeset and remove the bulk changeset, it doesn't make sense in general to have a user without role, they won't have permission for anything

ghatighorias: well, I think we should do it in another PR. 12 test from other controllers will break

ghatighorias: I opened a pr for adding role to user changeset "https://github.com/digitalnatives/course_planner/pull/225"

ggpasqualino: I merged it, you may use that already

ghatighorias: done

test/controllers/bulk_controller_test.exs
```diff
@@ -0,0 +1,197 @@
...
+    user = insert(role)
+
+    conn =
+      Phoenix.ConnTest.build_conn()
+      |> assign(:current_user, user)
+    {:ok, conn: conn}
+  end
+
+  defp create_input_params(target, title, csv_data) do
+    File.touch!(@file_path)
```
ggpasqualino: I think it would be better to create the file with `Plug.Upload.random_file!/1`
This creates the file in a proper temp place and removes it after the process dies

something like
```elixir
upload = Plug.Upload.random_file!("csv")
File.write!(upload.path, csv_data)
 %{"input" => %{"csv_file" => upload, "target" => target, "title" => title}}
```

ghatighorias: Done, but I had to modify it a bit cause just return a path.

ggpasqualino: cool! I confused it, I thought it would return a `%Plug.Upload{}` struct

PR #233, opened 2017-09-14
restrict the user list to be clickable only for coordinators

ghatighorias:
description : https://app.nostromo.io/work/course-planner/students-and-teachers-redirects-to-403-when-clicking-on-classmate-or-teachers-profile?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjk2Mjk5LGNhcmRfdGl0bGU6c3R1ZGVudHMtYW5kLXRlYWNoZXJzLXJlZGlyZWN0cy10by00MDMtd2hlbi1jbGlja2luZy1vbi1jbGFzc21hdGUtb3ItdGVhY2hlcnMtcHJvZmlsZX0=

### DIFF
--- a/lib/course_planner_web/views/shared_view.ex
+++ b/lib/course_planner_web/views/shared_view.ex
@@ -283,13 +283,15 @@ defmodule CoursePlannerWeb.SharedView do
 
   # show pages
 
-  def user_list(users, opts \\ []) do
+  def user_list(current_user, users, opts \\ []) do
+    clickable = current_user.role == "Coordinator"
     empty_text = opts[:empty_text] || "There are no users here yet"
     render "user_list.html", users: users,
-                             empty_text: empty_text
+                             empty_text: empty_text,
+                             clickable: clickable
   end
 
-  def user_bubble(user) do
+  def user_bubble(user, clickable) do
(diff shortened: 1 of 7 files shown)


lib/course_planner_web/views/shared_view.ex
```diff
@@ -283,13 +283,15 @@ defmodule CoursePlannerWeb.SharedView do
...
+    clickable = current_user.role == "Coordinator"
     empty_text = opts[:empty_text] || "There are no users here yet"
     render "user_list.html", users: users,
-                             empty_text: empty_text
+                             empty_text: empty_text,
+                             clickable: clickable
   end
 
-  def user_bubble(user) do
+  def user_bubble(user, clickable) do
```
ggpasqualino: what if we don't change the `lib/course_planner_web/templates/shared/user_bubble.html.eex` template, and we change the url instead to blank if it is not clickable?

ghatighorias: so you mean i pass a empty url to the user_bubble.html

ghatighorias: done

PR #234, opened 2017-09-14
restricts self deletion for coordinators

ghatighorias:
description : https://app.nostromo.io/work/course-planner/self-deletion-can-cause-the-app-unstable?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjk2MzAwLGNhcmRfdGl0bGU6c2VsZi1kZWxldGlvbi1jYW4tY2F1c2UtdGhlLWFwcC11bnN0YWJsZX0=

### DIFF
--- a/lib/course_planner/accounts/users.ex
+++ b/lib/course_planner/accounts/users.ex
@@ -39,15 +39,15 @@ defmodule CoursePlanner.Accounts.Users do
     end
   end
 
-  def delete(id, current_user_id) do
+  def delete(id, current_user_id \\ "") do
     if to_string(id) == to_string(current_user_id) do
       {:error, :self_deletion}
     else
-      delete(id)
+      do_delete(id)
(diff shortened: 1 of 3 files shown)


lib/course_planner/accounts/users.ex
```diff
@@ -39,6 +39,14 @@ defmodule CoursePlanner.Accounts.Users do
...
 
+  def delete(id, current_user_id) do
+    if to_string(id) == to_string(current_user_id) do
+      {:error, :self_deletion}
+    else
+      delete(id)
+    end
+  end
+
   def delete(id) do
```
ggpasqualino: maybe it is better to make this one a private function called `do_delete` so we don't get confused later

ghatighorias: done

PR #241, opened 2017-09-15
Display times according to timezone

rhnonose:
- [x] Configure the timezone as a setting
- [x] Convert to configured timezone when displaying
- [x] Convert to UTC when storing


lib/course_planner_web/templates/task/form.html.eex
```diff
@@ -12,8 +12,8 @@
               </div>
             <% end %>
             <%= CoursePlannerWeb.SharedView.form_text f, :name, required: true %>
-            <%= CoursePlannerWeb.SharedView.form_datetime f, :start_time, required: true, value: %{hour: 8, min: 0} %>
```
ghatighorias: why removing of the value ?

priv/repo/seeds.exs
```diff
@@ -159,4 +159,16 @@ unless Repo.get_by(SystemVariable, key: "NOTIFICATION_JOB_EXECUTED_AT") do
       }) |> Repo.insert!
 end
 
+unless Repo.get_by(SystemVariable, key: "TIMEZONE") do
```
ghatighorias: nice idea

lib/course_planner_web/templates/task/form.html.eex
rhnonose: The view's function puts the current time as a default, I thought it was supposed to be like that before but it wasn't exactly.

ghatighorias: in case you have checked that it puts it as default then ignore this comment

lib/course_planner/tasks/task.ex
```diff
@@ -30,6 +30,8 @@ defmodule CoursePlanner.Tasks.Task do
     |> validate_required(@required_params)
     |> validate_datetime()
     |> validate_expiration()
+    |> convert_timezone(:start_time)
```
ghatighorias: why the input time by the user needs convert ? cause user always put her time according to her own local timezone

lib/course_planner_web/templates/task/show.html.eex
```diff
@@ -33,15 +33,15 @@
                 <div class="detail">
                   <div class="detail__name">Start time</div>
                   <div class="detail__value">
-                    <%= String.slice NaiveDateTime.to_iso8601(@task.start_time), 0, 10 %>
-                    <%= String.slice NaiveDateTime.to_iso8601(@task.start_time), 11, 5 %>
+                    <%= String.slice NaiveDateTime.to_iso8601(@task.start_time |> Settings.utc_to_system_timezone()), 0, 10 %>
```
ghatighorias: as mentioned above, time set by user is always set according to her own timezone. we don't need conversion on it. or unless it will be double conversion

ghatighorias:
I think rather timezone should be only applied to Timex.now(), and not when user directly put a time and system save it. can you please check this scenarios ?

lib/course_planner/tasks/task.ex
rhnonose: Because we want to store it in UTC
Say the user inserts a 14:00 in her timezone and stores it naively (not converting)
If we compare anywhere to a time generated by say `Timex.now`, it's gonna be "off".

lib/course_planner_web/templates/task/show.html.eex
rhnonose: It should be converted to UTC to store and converted back to the configured timezone when displaying

rhnonose:
@ghatighorias there's an inherent problem in the way you mentioned. If we store naively, we can't for example change the timezone configuration without doing a massive update in the database (sucks), we also have to convert every time we want to compare to other times.
To deal with cases where the localtime changes (like summer time) is really hard.
It's also hard to distinguish if it's local or utc (since a lot of `datetime` types are not `utc_datetime` and we can still store in utc using `naive_datetime` "by assumption").

ghatighorias:
i approved this and let's test it cause i am a bit lost with this pr

rhnonose:
Me too, it's been a while.

PR #244, opened 2017-09-19
increases the test coverage

ghatighorias:


test/lib/course_planner/accounts/user_test.exs
```diff
@@ -105,4 +137,18 @@ defmodule CoursePlanner.UserTest do
       assert changeset.errors[:notification_period_days] == {"must be less than or equal to %{number}", [validation: :number, number: 7]}
     end
   end
+
+  describe "password changeset" do
+    test "when setting password" do
+      coordinator = insert(:coordinator)
+      changeset = User.changeset(coordinator, %{password: "secret", password_confirmation: "secret"}, :password)
+      assert [] == changeset.errors
```
rhnonose: Is there a reason for not using `assert changese.valid?`?

test/lib/course_planner/classes/classes_test.exs
```diff
@@ -0,0 +1,119 @@
...
+  test "split past and next classes for multiple classes" do
+    class1 = insert(:class, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: -2))
+    class2 = insert(:class, date: Timex.now(), starting_at: Timex.shift(Timex.now(), hours: +1))
+    class3 = insert(:class, date: Timex.now(), starting_at: Timex.now())
+    class4 = insert(:class, date: Timex.now(), starting_at: Timex.shift(Timex.now(), minutes: -2))
+    class5 = insert(:class, date: Timex.now(), starting_at: Timex.shift(Timex.now(), minutes: +2))
+
+    classes = [class1, class2, class3, class4, class5]
+    expected_past_classes =
+      [class1.id, class4.id]
```
rhnonose: In the previous test you asserted the whole class struct, but here you're asserting only the id. Was there a problem asserting the whole struct? Would be way less verbose.

rhnonose:
Great job.

test/lib/course_planner/classes/classes_test.exs
ghatighorias: i check this out , cant remember :P

test/lib/course_planner/accounts/user_test.exs
ghatighorias: no there is no reason, i think i was just lost

test/lib/course_planner/classes/classes_test.exs
ghatighorias: the reason is that in the previous test is testing a sorting function and this test just give past and previous classes so it might not be sorted, then I had to sort them based on id so they are going to be stable.
but now i sort them and check the whole struct. I see the point you are mentioning as well

test/lib/course_planner/accounts/user_test.exs
ghatighorias: done

test/lib/course_planner/classes/classes_test.exs
ghatighorias: done

PR #246, opened 2017-09-19
Makes login case insensitive

ghatighorias:
Check session_controller.ex line 81

And yes, for this very line i had to regenerate all the coherence controllers

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -49,7 +49,7 @@ defmodule CoursePlanner.Mixfile do
      {:gettext, "~> 0.11"},
      {:cowboy, "~> 1.0"},
      {:swoosh, "~> 0.7.0"},
-     {:coherence, "0.5.0"},
+     {:coherence, github: "ghatighorias/coherence", tag: "v0.5.1"},
      {:canada, "~> 1.0.1"},
      {:canary, github: "cpjk/canary"},
      {:distillery, "~> 1.0.0"},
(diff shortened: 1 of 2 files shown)


ghatighorias:
@rhnonose @ggpasqualino shall we cover this coherence files ? :( :( 
can't we exclude them from the coveralls

rhnonose:
Hmm, maybe we copy coherence's tests? 😄 
I think we should ignore, but there's probably a better way.

ggpasqualino:
https://github.com/digitalnatives/course_planner/pull/246#issuecomment-330528095

we can test it or not, excluding them is not an option in my opinion, files should only be excluded from the coverage if it doesn't make sense to test them, the files inside the the `test/support` for example

rhnonose:
Make it a PR to generate the tests together with the code, maybe?

ghatighorias:
@rhnonose  I don't understand

rhnonose:
@ghatighorias Weren't the files generated by coherence?

ghatighorias:
I am sorry @rhnonose but that will take ages, he is not responsive and apparently the controller generation function is totally changed in the new version that is about to relase

rhnonose:
@ghatighorias I guess wen can just keep it in mind that we can use the tests in a PR later, doesn't have to be integrated first since it's a generated code.

ghatighorias:
importing tests from coherence is hopeless. they all fail and needs to be rewritten. it's not all. the problem is that he only wrote tests for Trackable user ( one which we don't use in our app at all ).

what if I fork it, fix the issue in the fork and use my fork of coherence ? this way we don't need to deal with the coverage and test and so on

ggpasqualino:
> what if I fork it, fix the issue in the fork and use my fork of coherence ? this way we don't need to deal with the coverage and test and so on

@ghatighorias I think that's probably better

PR #249, opened 2017-09-20
Ghatighorias/organize offered course index page

ghatighorias:


test/lib/course_planner/courses/offered_courses_test.exs
```diff
@@ -45,33 +45,39 @@ defmodule CoursePlanner.OfferedCoursesTest do
   describe "find_all_by_user/1" do
     test "coordinators should see all offered courses" do
       user = insert(:coordinator)
-      insert_list(2, :offered_course)
+      term = insert(:term)
+      insert_list(2, :offered_course, term: term)
 
-      courses = OfferedCourses.find_all_by_user(user)
+      terms = OfferedCourses.find_all_by_user(user)
```
ggpasqualino: this became awkward, `OfferedCourses.find_all_by_user` doesn't return offered courses, but terms
I don't think this interface should change

ghatighorias: shall i change it to something like "find_all_by_user_categorized_by_term" ?
or shall I move it to terms.ex ?

ggpasqualino: I think it would make more sense to move it to terms

ghatighorias: done

PR #251, opened 2017-09-21
Fixes wrong attendance order

ghatighorias:
ordering was forgotten to be added for the teacher and was just added for the coordinator. this commit ensures the ordering based on the student name and family_name to be added for the teacher as well

root cause :
in the show we use "get_teacher_course_attendances" while in fill we use "get_course_attendances" function to retrive all the attendances. both have to have the same sorting order so the rendered order of the students and attendance be the same

The same problem won't happen for student cause he only sees his own attendance and need no sorting on name or family_name

### DIFF
--- a/lib/course_planner/attendances/attendances.ex
+++ b/lib/course_planner/attendances/attendances.ex
@@ -31,7 +31,7 @@ defmodule CoursePlanner.Attendances do
       join: a in assoc(c,  :attendances),
       join: as in assoc(a, :student),
       join: ac in assoc(a, :class),
-      preload: [:term, :course, teachers: s, students: t],
+      preload: [:term, :course, teachers: t, students: s],
       preload: [classes: {c, students: cs, attendances: {a, student: as, class: ac}}],
       where: oc.id == ^offered_course_id and t.id == ^teacher_id,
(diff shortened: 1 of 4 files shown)


lib/course_planner/attendances/attendances.ex
```diff
@@ -27,10 +27,12 @@ defmodule CoursePlanner.Attendances do
       join: s in assoc(oc, :students),
       join: c in assoc(oc, :classes),
       join: a in assoc(c,  :attendances),
+      join: as in assoc(a, :student),
```
ggpasqualino: I don't get why these two new joins are necessary

ghatighorias: we have had this problem with "get_course_attendances" and fixed it with this joins and sort them
in the show and fill I use offered_course.students to create a part of the table but then iterate throw offered_course.class.attendance ( which has student link too ) to check the attendance of each student, so both should be sorted the same in order columns and attendance data to match

ghatighorias:
even this solution is not complete, adding some more student makes it completely messy

ghatighorias:
The last comment fixes all the issues, the ordering should have been done from the beginning based on attendance student and not offered_course student and likewise drawing the tables in the show and fill pages

PR #254, opened 2017-09-25
Use phoenix 1.3's action_fallback

rhnonose:
- Use action_fallback to extract common controller logic
- Move logic out of controllers


ghatighorias:
@rhnonose this will take me another day to finish, it is a massive change so if you don't mind i'll be slower in order to check everything as good as i can

rhnonose:
@ghatighorias did you review this?
It's been here a while, we should merge.

PR #259, opened 2017-09-28
Properly check e-mail format.

rhnonose:
Properly check e-mail format using `email_checker`.


ghatighorias:
Just a question. does the email checker accept emails in the following format as well 
myname.myfamily@maindomain.subsection.subsubsection

( cause it is very often the case) like for olive program is like this "name.family@student.ceu.edu"

rhnonose:
@ghatighorias both mentioned e-mails are valid

PR #260, opened 2017-10-04
Adds student attendances to show page

ghatighorias:
This PR give the student show page an overview of her attendance to all classes grouped by terms


ghatighorias:
Where condition is changed to user student from attendance. the reason is that a class has many attendance and filtering based on offered_course.student_id does not mean that the correct attendance is loaded too. it just mean the first attendance is loaded

PR #265, opened 2017-10-19
makes some other program settings as text type

ghatighorias:
this PR is for covering the following card 
1.
https://app.nostromo.io/work/course-planner/text-stylings-like-newline-is-ignored-in-text-and-string-fields?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjk2NDYzLGNhcmRfdGl0bGU6dGV4dC1zdHlsaW5ncy1saWtlLW5ld2xpbmUtaXMtaWdub3JlZC1pbi10ZXh0LWFuZC1zdHJpbmctZmllbGRzfQ==

following fields were forgotten in the previous or that are now covered "address, phone number"
As for website since it is a url field it cannot be multi-lined and as for this pr it won't be covered and will be removed from the card's scope

2. seeds could not be updated when they were once created, I have change the way so if the settings are created  we be able still to update some parts of their variables

3. updates the user comment to text field to allow longer comment be given for a student. makes it to support multilined


rhnonose:
Is 1. missing? Only `seeds.exs` is modified.

ghatighorias:
Good point. I update it. sorry

PR #270, opened 2017-10-24
Events page

rhnonose:
- Add pages for events (index, show, new, edit, delete)
- Students, teachers and volunteers can only list and see events
- Moved controller tests to the correct location


lib/course_planner_web/templates/event/form.html.eex
```diff
@@ -0,0 +1,32 @@
...
+                <p>Oops, something went wrong! Please check the errors below.</p>
+              </div>
+            <% end %>
+            <%= CoursePlannerWeb.SharedView.form_text f, :name, required: true %>
+            <%= CoursePlannerWeb.SharedView.form_text f, :location, required: true %>
+            <%= CoursePlannerWeb.SharedView.form_date f, :date, required: true %>
+            <%= CoursePlannerWeb.SharedView.form_time f, :starting_time, required: true, value: %{hour: 8, min: 0}  %>
+            <%= CoursePlannerWeb.SharedView.form_time f, :finishing_time, required: true, value: %{hour: 8, min: 0}  %>
+            <%= CoursePlannerWeb.SharedView.form_textarea f, :description %>
+            <%= CoursePlannerWeb.SharedView.form_multiselect f, :user_ids, all_users(), selected: selected_users(@changeset),
```
ghatighorias: this really needs a Add_all functionality. I know this does not fall in this PR. but seriously creating an event for many users is gonna be a pain in the ass

lib/course_planner_web/controllers/event_controller.ex
```diff
@@ -0,0 +1,76 @@
...
+  use CoursePlannerWeb, :controller
+
+  alias CoursePlanner.Events
+  alias CoursePlanner.Events.Event
+
+  import Canary.Plugs
+  plug :authorize_controller
+
+  def index(conn, _params) do
+    events = Events.all()
```
ghatighorias: can you sort this in descending order, so the latest event be on top?
this way it's more user friendly

ghatighorias: plus maybe we actually should do what is already is in task index page. to have past event, and upcoming separated ? what do you think ?

lib/course_planner_web/controllers/permissions.ex
```diff
@@ -36,5 +37,10 @@ defimpl Canada.Can, for: CoursePlanner.Accounts.User do
   def can?(%User{id: id}, action, %User{id: id})
     when action in [:show, :edit, :update], do: true
 
+  def can?(_user, action, EventController)
+    when action in [:index, :show], do: true
+
+  def can?(_user, _action, EventController), do: false
```
ghatighorias: isn't this line already covered in line 45 ?
`def can?(_user, _action, _controller), do: false`

test/course_planner_web/controllers/event_controller_test.exs
```diff
@@ -0,0 +1,206 @@
...
+    end
+
+    @tag user_role: :coordinator
+    test "deletes chosen event", %{conn: conn, event: event} do
+      conn = delete conn, event_path(conn, :delete, event)
+      assert redirected_to(conn) == event_path(conn, :index)
+      conn = get conn, event_path(conn, :show, event)
+      response(conn, 404)
+    end
+  end
```
ghatighorias: you do not test deletion of a non existing resource when user is not permitted. the point of testing it is to see which error will be visible. resource does not exist or action is not permitted.

lib/course_planner_web/templates/event/form.html.eex
rhnonose: We definitely do.

lib/course_planner_web/controllers/permissions.ex
rhnonose: Yeah it is. I forgot it's a whitelist approach now.

test/course_planner_web/controllers/event_controller_test.exs
rhnonose: Good catch, I'll add that.

lib/course_planner_web/controllers/event_controller.ex
rhnonose: Probably.
I'll leave both to a different PR since I'll have to undo/redo/do differently the sorting.

ghatighorias: then lets have a card for it so we don't forget

PR #271, opened 2017-10-24
logging in through json

ghatighorias:
initial draft

### DIFF
--- a/lib/course_planner_web/controllers/auth/api_session_controller.ex
+++ b/lib/course_planner_web/controllers/auth/api_session_controller.ex
@@ -1,0 +1,34 @@
+defmodule CoursePlannerWeb.Auth.Api.V1.SessionController do
+  @moduledoc """
+    This module handles api loging in to the system
+  """
+  use CoursePlannerWeb, :controller
+
+  alias CoursePlanner.Accounts.{Users, User}
+
+  plug :put_layout, ""
+
+  def create(conn, %{"email" => email, "password" => password}) do
+    trimmed_downcased_email =
+      email
+      |> String.trim()
+      |> String.downcase()
+    user = Repo.get_by(User, email: trimmed_downcased_email)
+
+    case Users.check_password(user, password) do
+      {:ok, _reason} ->
+        conn
+        |> json(%{ token:  get_login_token(user) })
+
+      {:error, _reason} ->
+        conn
+        |> json(%{ token:  "error" })
+    end
+  end
+
+  defp get_login_token(user) do
+    { :ok, jwt, _full_claims } = Guardian.encode_and_sign(user, :api)
+    jwt
+  end
+
+end
--- a/lib/course_planner_web/controllers/auth/api_session_controller.ex
+++ b/lib/course_planner_web/controllers/auth/api_session_controller.ex
@@ -29,7 +29,7 @@ defmodule CoursePlannerWeb.Auth.Api.V1.JsonSessionController do
   defp get_login_token(conn, user) do
     conn
     |> Guardian.Plug.api_sign_in(user)
-    |> Guardian.Plug.current_token(new_conn)
+    |> Guardian.Plug.current_token()
   end
 
 end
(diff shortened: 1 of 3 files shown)


lib/course_planner_web/router.ex
```diff
@@ -29,6 +35,13 @@ defmodule CoursePlannerWeb.Router do
   end
 
   scope "/" do
+    pipe_through :public_api
+
+    resources "/api/sessions", CoursePlannerWeb.Auth.Api.V1.JsonSessionController,
```
rhnonose: The API versioning should be exposed to the client, so the `v1` should be either in the `scope` or in the route.

lib/course_planner_web/controllers/auth/api_session_controller.ex
```diff
@@ -0,0 +1,41 @@
+defmodule CoursePlannerWeb.Auth.Api.V1.JsonSessionController do
```
rhnonose: Not sure about the `V1` in the module, the nesting and the directory structure doesn't match. Can you fix that?

lib/course_planner_web/controllers/auth/api_session_controller.ex
```diff
@@ -0,0 +1,41 @@
...
+  end
+  def create(conn, _params) do
+    conn
+    |> put_status(406)
+    |> render(CoursePlannerWeb.ErrorView, "406.json", errors: [])
+  end
+
+  defp get_login_token(conn, user) do
+    conn
+    |> Plug.api_sign_in(user)
```
rhnonose: Since now there's gonna be a public API for login, maybe we need the ability to revoke tokens.
Thoughts on that? Will need to set something like `GuardianDb`.

lib/course_planner_web/controllers/auth/api_session_controller.ex
ghatighorias: What do you suggest cause i have no idea, i just saw this as a suggestion. 
1 - create the folder structure
2 - remove V1

lib/course_planner_web/controllers/auth/api_session_controller.ex
ghatighorias: Agreed.
As I have seen this is a functionality provided by guardianDB.
So what do you suggest ? to have it under this scope or another PR ?

lib/course_planner_web/router.ex
ghatighorias: if we agree to have the V1 in the name of the module then i suggest to have it as scope

lib/course_planner_web/controllers/auth/api_session_controller.ex
rhnonose: Another PR. We should see if we'll need the revoke functionality.

lib/course_planner_web/controllers/auth/api_session_controller.ex
rhnonose: For now we can remove the V1 from the module. Internally we can rename/move anytime, we should only leave the route consistent.

lib/course_planner_web/router.ex
rhnonose: Let's remove the module name an have the `/api/v1` on the scope.

ghatighorias: done

lib/course_planner_web/controllers/auth/api_session_controller.ex
ghatighorias: done

lib/course_planner_web/controllers/auth/api_session_controller.ex
ghatighorias: Card added
https://app.nostromo.io/work/course-planner/investigate-and-add-revoke-functionality-for-login-tokens-is-decided?e21vZHVsZTp3b3JrLHByb2plY3Q6NjM2OCxjYXJkOjk4MDIzLGNhcmRfdGl0bGU6aW52ZXN0aWdhdGUtYW5kLWFkZC1yZXZva2UtZnVuY3Rpb25hbGl0eS1mb3ItbG9naW4tdG9rZW5zLWlzLWRlY2lkZWR9

PR #273, opened 2017-10-26
Event nofitications

rhnonose:
Notify users by e-mail when:
- An event is created with the users added
- An event is updated
  - User is removed (uninvited)
  - User is added (invited)
  - User stays (event is updated)
- An event is deleted


lib/course_planner/events/event_encoder.ex
```diff
@@ -0,0 +1,19 @@
+defimpl Poison.Encoder, for: CoursePlanner.Events.Event do
```
ghatighorias: what is the user of this code? i am not familiar

rhnonose: It's an Elixir Protocol. `defimpl` overrides the standard encoder (that tries to encode everything inside the `Event` struct). I was having some trouble with the `users` relationship.

PR #278, opened 2017-10-30
Split event in index by past and future.

rhnonose:
Also, fix event sorting:
- was only considering `starting_time` and `finishing_time` since the `date` part is separated
- was `desc` instead of `asc`

### DIFF
--- a/lib/course_planner/events/events.ex
+++ b/lib/course_planner/events/events.ex
@@ -40,7 +40,7 @@ defmodule CoursePlanner.Events do
   def all_with_users do
     query = from e in Event,
     preload: [:users],
-    order_by: [desc: e.date, desc: e.starting_time, desc: e.finishing_time]
+    order_by: [asc: e.date, asc: e.starting_time, asc: e.finishing_time]
 
     Repo.all(query)
   end
--- a/lib/course_planner/events/events.ex
+++ b/lib/course_planner/events/events.ex
@@ -30,7 +30,7 @@ defmodule CoursePlanner.Events do
 
   def compare_event_date_time(event, now) do
     {:ok, event_datetime} = NaiveDateTime.new(event.date, event.starting_time)
-    Timex.compare(event_datetime, now) == -1
+    Timex.before?(event_datetime, now)
   end
 
   def revertse_past_events({past_events, future_events}) do
--- a/lib/course_planner/events/events.ex
+++ b/lib/course_planner/events/events.ex
@@ -17,15 +17,30 @@ defmodule CoursePlanner.Events do
 
   def all do
     query = from e in Event,
-    order_by: [desc: e.starting_time, desc: e.finishing_time]
+    order_by: [asc: e.date, asc: e.starting_time, asc: e.finishing_time]
 
     Repo.all(query)
+  end
+
+  def all_splitted(now) do
+    all()
+    |> Enum.split_with(&(compare_event_date_time(&1, now)))
+    |> revertse_past_events()
+  end
+
+  def compare_event_date_time(event, now) do
+    {:ok, event_datetime} = NaiveDateTime.new(event.date, event.starting_time)
+    Timex.compare(event_datetime, now) == -1
+  end
+
+  def revertse_past_events({past_events, future_events}) do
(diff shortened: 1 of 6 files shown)


lib/course_planner/events/events.ex
```diff
@@ -17,15 +17,30 @@ defmodule CoursePlanner.Events do
 
   def all do
     query = from e in Event,
-    order_by: [desc: e.starting_time, desc: e.finishing_time]
+    order_by: [asc: e.date, asc: e.starting_time, asc: e.finishing_time]
```
ghatighorias: why did you change the ordering to ascending ?

lib/course_planner/events/events.ex
```diff
@@ -17,15 +17,30 @@ defmodule CoursePlanner.Events do
...
 
     Repo.all(query)
   end
 
+  def all_splitted(now) do
```
ghatighorias: is there any logic to distinguish between all and the events for a specific person ?
I think again like courses in the index page we should not show other events to people who are not added as participants, but rather provide the option via calendar using "my_classes" option

test/course_planner/events/events_test.exs
```diff
@@ -21,6 +21,20 @@ defmodule CoursePlanner.EventsTest do
       assert Events.all_with_users() == [event |> Repo.preload(:users)]
     end
 
+    test "all_splitted/1 returns events splited by time" do
+      %{id: id1} = insert(:event, %{date: ~D[2017-02-02], starting_time: ~T[10:00:00.000000]})
```
ghatighorias: please use Timex for setting time in future or past or your test becomes obsolete after a while
E.g. `Timex.shift(Timex.now(), days: -10))`

lib/course_planner/events/events.ex
rhnonose: No, it's not. It was created as a default index and I understood that all the events would be visible, but I guess I was wrong.

lib/course_planner/events/events.ex
rhnonose: I'm not sure, I had the impression that `desc` was wrong.

ghatighorias: This way the earliest event ever to happen always stay on top regardless of the current date, so after create some events, the user should always scroll down.

I have a few suggestion, but do tell me what do you think
1 - to have three sections : (past and upcoming events) . now if past events are sorted Descending and upcoming is sorted ascending the one on the top is whether the one that lately happen or the one that is going to happen the earliest  
2 - to sort everything descending, this i have applied in other places and solve the problem for a while
3 - to write a generic sort solution for events, task, terms and classes. mainly for all entities that have time or date or both

rhnonose: We can finish this PR with **2** and discuss **3**

lib/course_planner_web/templates/event/index.html.eex
```diff
@@ -10,74 +10,7 @@
...
-                  <%= link "Delete", to: event_path(@conn, :delete, event), method: :delete, data: [confirm: "Are you sure?"] %>
-                </li>
-              <% end %>
-            </ul>
-          </td>
-        </tr>
-      <% end) %>
-    </tbody>
-  </table>
+<%= render "index_event.html", events: @future_events, section_title: "Future events", conn: @conn %>
```
ghatighorias: what do you think calling it "upcoming events". just a suggestion and i won't block the merge for this indeed

lib/course_planner/events/events.ex
```diff
@@ -17,15 +17,30 @@ defmodule CoursePlanner.Events do
...
+    |> Enum.split_with(&(compare_event_date_time(&1, now)))
+    |> revertse_past_events()
+  end
+
+  def compare_event_date_time(event, now) do
+    {:ok, event_datetime} = NaiveDateTime.new(event.date, event.starting_time)
+    Timex.before?(event_datetime, now)
+  end
+
+  def revertse_past_events({past_events, future_events}) do
```
ghatighorias: "revertse" is typo ?

PR #279, opened 2017-10-31
Supervisor role

rhnonose:
- [x] Add supervisor as a role
- [x] Grant read-only permission to everything
- [x] Increment tests


lib/course_planner_web/templates/layout/supervisor_app_navbar.html.eex
```diff
@@ -0,0 +1,23 @@
+<%= CoursePlannerWeb.SharedView.navbar "Course Planner" do %>
```
ghatighorias: is there a difference between coordinator and supervisor's navbar ?

ghatighorias:
Just a point. i am not sure if it's needed for supervisor to get notification as well. cause it's rather looking at things. what do you think ?

rhnonose:
@ghatighorias Since there's the option to have notification in x period of days, maybe we can add an option to disable it (or usa 0).

lib/course_planner_web/templates/layout/supervisor_app_navbar.html.eex
rhnonose: Nope, though I might need to customize it so made a copy.

PR #284, opened 2017-11-06
Fixes password length check when user updates itself

ghatighorias:
password limit of 8 character was just being checked on password reset page and not on user update page ( it was wrongly still on 6 characters )

### DIFF
--- a/test/course_planner_web/controllers/user_controller_test.exs
+++ b/test/course_planner_web/controllers/user_controller_test.exs
@@ -206,6 +206,15 @@ defmodule CoursePlanner.UserControllerTest do
     assert Repo.get_by(User, email: "foo@bar.com")
   end
 
+  test "does not update password if is too short" do
+    user = insert(:student, name: "Foo", family_name: "Bar")
+    user_conn = guardian_login_html(user)
+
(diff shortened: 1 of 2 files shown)


test/course_planner_web/controllers/user_controller_test.exs
```diff
@@ -206,6 +206,15 @@ defmodule CoursePlanner.UserControllerTest do
     assert Repo.get_by(User, email: "foo@bar.com")
   end
 
+  test "does not update password if is too short" do
+    user = insert(:student, name: "Foo", family_name: "Bar")
+    user_conn = guardian_login_html(user)
```
rhnonose: You can simply `login_as(:student)`, it'll churn some code.

ghatighorias: yes but if you see i need the user as well. then afterward i have get the user from conn which makes it the same two lines :)

rhnonose: Well, you can `%{current_user: user} = conn = login_as(:student)`.
I'll approve regardless since I'm being picky 😄

ghatighorias: nooice. sorry didn't see this coming :D

ghatighorias: done

